PR #6, opened 2022-11-08
ws-protocol server, ROS2 msgdef lookup, ROS2 foxglove_bridge implementation

jhurliman:
**Public-Facing Changes**
- ROS 2 node starts a WebSocket server speaking `ws-protocol`

**Description**
This brings the ROS 2 side of things closer to "feature complete" for read-only topic subscriptions for `.msg`-backed files. Not supported:

- All QoS profiles
- IDL message definitions
- Parameters
- Services
- Actions
- Publishing
- Non-CDR serializations
- ROS 1 node

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,3 +1,4 @@
+// -*- jsonc -*-
 {
   // Use IntelliSense to learn about possible attributes.
   // Hover to view descriptions of existing attributes.
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,18 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [{
+    "name": "C++ Debug",
+    "type": "cppdbg",
+    "request": "launch",
+    "program": "${workspaceFolder}/build/foxglove_bridge/foxglove_bridge",
+    "args": [],
+    "stopAtEntry": false,
+    "cwd": "${workspaceFolder}",
+    "environment": [],
+    "externalConsole": false,
+    "MIMode": "gdb",
+  }]
+}
--- a/foxglove_bridge_base/include/foxglove_bridge/message_definition_cache.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/message_definition_cache.hpp
@@ -1,0 +1,79 @@
+#pragma once
+
+#include <set>
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+
+namespace foxglove {
+
+enum struct MessageDefinitionFormat {
+  IDL,
+  MSG,
+};
+
+struct MessageSpec {
+  MessageSpec(MessageDefinitionFormat format, std::string text, const std::string& package_context);
+  const std::set<std::string> dependencies;
+  const std::string text;
+  MessageDefinitionFormat format;
+};
+
+struct DefinitionIdentifier {
+  MessageDefinitionFormat format;
+  std::string package_resource_name;
+
+  bool operator==(const DefinitionIdentifier& di) const {
+    return (format == di.format) && (package_resource_name == di.package_resource_name);
+  }
+};
+
+class DefinitionNotFoundError : public std::exception {
+private:
+  std::string name_;
+
+public:
+  explicit DefinitionNotFoundError(std::string name)
+      : name_(std::move(name)) {}
+
+  const char* what() const noexcept override {
+    return name_.c_str();
+  }
+};
+
+class MessageDefinitionCache final {
+public:
+  /**
+   * Concatenate the message definition with its dependencies into a self-contained schema.
+   * The format is different for MSG and IDL definitions, and is described fully here:
+   * [MSG](https://mcap.dev/specification/appendix.html#ros2msg-data-format)
+   * [IDL](https://mcap.dev/specification/appendix.html#ros2idl-data-format)
+   * Throws DefinitionNotFoundError if one or more definition files are missing for the given
+   * package resource name.
+   */
+  std::pair<MessageDefinitionFormat, std::string> get_full_text(
+    const std::string& package_resource_name);
+
+private:
+  struct DefinitionIdentifierHash {
+    std::size_t operator()(const DefinitionIdentifier& di) const {
+      std::size_t h1 = std::hash<MessageDefinitionFormat>()(di.format);
+      std::size_t h2 = std::hash<std::string>()(di.package_resource_name);
+      return h1 ^ h2;
+    }
+  };
+  /**
+   * Load and parse the message file referenced by the given datatype, or return it from
+   * msg_specs_by_datatype
+   */
+  const MessageSpec& load_message_spec(const DefinitionIdentifier& definition_identifier);
+
+  std::unordered_map<DefinitionIdentifier, MessageSpec, DefinitionIdentifierHash>
+    msg_specs_by_definition_identifier_;
+};
+
+std::set<std::string> parse_dependencies(MessageDefinitionFormat format, const std::string& text,
+                                         const std::string& package_context);
+
+}  // namespace foxglove
--- a/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
@@ -24,7 +24,6 @@ using json = nlohmann::json;
 using namespace std::placeholders;
 
 using ConnHandle = websocketpp::connection_hdl;
-using MessagePtr = WebSocketNoTlsServer::message_ptr;
 using OpCode = websocketpp::frame::opcode::value;
 
 using ChannelId = uint32_t;
@@ -77,8 +76,13 @@ enum class StatusLevel : uint8_t {
   ERROR = 2,
 };
 
+template <typename ServerConfiguration>
 class Server final {
 public:
+  using ServerType = websocketpp::server<ServerConfiguration>;
+  using ConnectionType = websocketpp::connection<ServerConfiguration>;
+  using MessagePtr = typename ServerType::message_ptr;
+
   static const std::string SUPPORTED_SUBPROTOCOL;
 
   explicit Server(std::string name, LogCallback logger);
@@ -118,7 +122,7 @@ private:
 
   std::string _name;
   LogCallback _logger;
-  WebSocketNoTlsServer _server;
+  ServerType _server;
   std::unique_ptr<std::thread> _serverThread;
 
   uint32_t _nextChannelId = 0;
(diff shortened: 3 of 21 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
+{
```
jtbandes, line 1:
```suggestion
// -*- jsonc -*-
{
```

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -0,0 +1,541 @@
+#pragma once
```
jtbandes, line 1: 🙁  can this file come from foxglove-websocket on conan instead of being vendored here? What were the changes needed?

jtbandes: I guess we are going to run into the same packaging issues we did with mcap that led to mcap_vendor. Time for a foxglove_websocket_vendor?

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -1,24 +1,343 @@
...
+    _server.sendMessage(channel.id, timestamp, payload);
+  }
 };
 
 int main(int argc, char* argv[]) {
   rclcpp::init(argc, argv);
-  rclcpp::spin(std::make_shared<FoxgloveBridge>());
+  auto bridgeNode = std::make_shared<FoxgloveBridge>();
+  rclcpp::spin(bridgeNode);
+  bridgeNode.reset();
```
jtbandes, line 340: this change shouldn't change anything?

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
jhurliman: Not really, ROS (1 or 2) build systems don't include conan. The closest we could do is create a foxglove_websocket_vendor package similar to what we did with mcap_vendor, but it would need to be done for both ROS 1 and ROS 2 and would really be overkill compared to this approach.

Aside from that, this is not the same implementation. `foxglove-websocket` blocks when you call `run()`, while this is a multi-threaded server that asynchronously runs the WebSocket event loop when you call `start()`.

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
jhurliman: It explicitly calls the `FoxgloveBridge` destructor to start the process of sending TCP connection close packets to connected clients as early as possible before we go into the ROS 2 shutdown. It might not make any difference, I need to test still.

jtbandes: If anything it would only _extend_ the lifetime of the FoxgloveBridge object (depending on whether spin() does anything to remove its strong reference before it returns)

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
jtbandes: What is our plan for dealing with bugfixes and protocol changes upstream in https://github.com/foxglove/ws-protocol/blob/main/cpp/foxglove-websocket/include/foxglove/websocket/server.hpp? Is that server implementation becoming abandonware? Are we maintaining two servers?

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -1,24 +1,367 @@
...
+
+  void subscribeHandler(foxglove::ChannelId channelId) {
+    std::lock_guard<std::mutex> lock(_subscriptionsMutex);
+    if (_subscriptions.find(channelId) != _subscriptions.end()) {
+      // Subscription already exists
+      RCLCPP_DEBUG(this->get_logger(), "Subscription already exists for channel %d", channelId);
+      return;
+    }
+
+    auto it = _channelToTopicAndDatatype.find(channelId);
```
achim-k, line 262:
```suggestion
    const auto it = _channelToTopicAndDatatype.find(channelId);
```
also in some places below

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -1,24 +1,367 @@
...
+    if (it == _channelToTopicAndDatatype.end()) {
+      RCLCPP_WARN(this->get_logger(), "Received subscribe request for unknown channel %d",
+                  channelId);
+      return;
+    }
+    auto& topicAndDatatype = it->second;
+    auto topic = topicAndDatatype.first;
+    auto datatype = topicAndDatatype.second;
```
achim-k, line 270:
```suggestion
    const auto& topicAndDatatype = it->second;
    const auto& [topic, datatype] = topicAndDatatype;
```

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -1,24 +1,367 @@
...
+                   channelId, topic.c_str(), datatype.c_str());
+      return;
+    }
+    auto& channel = it2->second;
+
+    rclcpp::SubscriptionEventCallbacks eventCallbacks;
+    eventCallbacks.incompatible_qos_callback = [&](const rclcpp::QOSRequestedIncompatibleQoSInfo&) {
+      RCLCPP_ERROR(this->get_logger(), "Incompatible subscriber QoS settings for topic \"%s\" (%s)",
+                   topic.c_str(), datatype.c_str());
+    };
```
achim-k, line 283: Shall we also send this error to the client (via the `status` operation)?

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -1,24 +1,367 @@
...
+
+    rclcpp::SubscriptionOptions subscriptionOptions;
+    subscriptionOptions.event_callbacks = eventCallbacks;
```
achim-k, line 286: We should also add a dedicated [callback_group](https://docs.ros2.org/galactic/api/rclcpp/structrclcpp_1_1SubscriptionOptionsBase.html#ab3998ee55c81d8035a0bd4b574200218) for each subscription. Otherwise all subscription share the default callback group and there is little multi threading (despite the multi threaded executor)

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -0,0 +1,571 @@
...
+  LogCallback _logger;
+  ServerType _server;
+  std::unique_ptr<std::thread> _serverThread;
+
+  uint32_t _nextChannelId = 0;
+  std::map<ConnHandle, ClientInfo, std::owner_less<>> _clients;
+  std::unordered_map<ChannelId, Channel> _channels;
+  std::function<void(ChannelId)> _subscribeHandler;
+  std::function<void(ChannelId)> _unsubscribeHandler;
+  std::shared_mutex _clientsChannelMutex;
```
achim-k, line 133: For the future it probably makes sense to have two separate mutexes, one for channels and one for clients.

foxglove_bridge_base/include/foxglove_bridge/message_definition_cache.hpp
```diff
@@ -0,0 +1,79 @@
...
+  struct DefinitionIdentifierHash {
+    std::size_t operator()(const DefinitionIdentifier& di) const {
+      std::size_t h1 = std::hash<MessageDefinitionFormat>()(di.format);
+      std::size_t h2 = std::hash<std::string>()(di.package_resource_name);
+      return h1 ^ h2;
+    }
+  };
+  /**
+   * Load and parse the message file referenced by the given datatype, or return it from
+   * msg_specs_by_datatype
```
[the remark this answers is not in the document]
jhurliman: This code was taken verbatim from [rosbag](https://github.com/ros-tooling/rosbag2_storage_mcap/blob/main/rosbag2_storage_mcap/include/rosbag2_storage_mcap/message_definition_cache.hpp) so I'm reticent to fork it.

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -1,24 +1,367 @@
...
+
+          // Add a mapping from the topic+datatype tuple to a dummy channel so we don't repeatedly
+          // try to load the message definition
+          auto channel = foxglove::Channel{0, newChannel};
+          _advertisedTopics.emplace(topicAndDatatype, std::move(channel));
+        }
+      }
+    }
+
+    if (newTopics.size() > 0) {
```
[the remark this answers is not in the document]
jhurliman: Not according to the current protocol implementation, where "advertise" is additive and you send "unadvertise" ops to remove channels.

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -1,24 +1,367 @@
...
+      _server.broadcastChannels();
+    }
+
+    // Schedule the next update
+    _updateCount++;
+    auto nextUpdateMs =
+      std::chrono::milliseconds(std::min(size_t(1) << _updateCount, _maxUpdateMs));
+    _updateTimer = this->create_wall_timer(
+      nextUpdateMs, std::bind(&FoxgloveBridge::updateAdvertisedTopics, this));
```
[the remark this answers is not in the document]
jhurliman: This is truncated exponential backoff, I'll add a comment. I think this is better than a constant period refresh since the bridge is coming online in parallel with the other nodes in the system, so this will result in faster end-to-end iteration speeds when you are frequently restarting the rosgraph.

Polling is annoying since it requires a separate thread and there is a balance between CPU usage and shutdown latency that has to be managed with the `timeout` argument. But it might still be better than this approach. Something to consider for a future PR.

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
jhurliman: Why?

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
jhurliman: No, this is internal to ROS 2 and not something a ws-protocol client has control over. We need to implement https://github.com/foxglove/ros-foxglove-bridge/issues/7 and then this would only happen in a rare circumstance where the publisher readvertises with a different QoS profile before the bridge opens a subscription.

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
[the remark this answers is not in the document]
jhurliman: https://github.com/foxglove/ros-foxglove-bridge/issues/15

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
jhurliman: https://github.com/foxglove/ros-foxglove-bridge/issues/16

PR #18, opened 2022-11-08
Hybrid ROS 1 / ROS 2 devcontainer

jhurliman:
**Public-Facing Changes**

**Description**
Use a single devcontainer.json that contains ROS 1 noetic and ROS 2 galactic in Ubuntu 20.04. Fixes https://github.com/foxglove/ros-foxglove-bridge/issues/17


achim-k (approved):
Tested it and works for me. When building for both ros1 and ros2, the `build` folder might collide, so one has to use e.g. `colcon build --build-base ros2_build` or `catkin config -b ros1_build`

PR #21, opened 2022-11-10
Add C++ library to get full message definition for ros1 types.

achim-k:
**Public-Facing Changes**
None

**Description**
For ROS1, the full ros message definition of a topic is only available when subscribing to the topic. In order to avoid that, a simple library was added that allows to generate the full message schema without having to subscribe to the topic. The library works similar to `roslib`'s `gendeps` tool in that it parses `.msg` file on the disk to generate a schema where all types are concatenated.

This library will be used to get the message schemas for published topics. These will then be advertised as available channels.

### DIFF
--- a/ros1_foxglove_bridge/include/foxglove_bridge/msg_parser.hpp
+++ b/ros1_foxglove_bridge/include/foxglove_bridge/msg_parser.hpp
@@ -1,9 +1,9 @@
 #pragma once
 
-#include <map>
 #include <memory>
 #include <stdexcept>
 #include <string>
+#include <unordered_map>
 #include <vector>
 
 namespace rospack {
@@ -43,7 +43,7 @@ private:
   std::string get_msg_file_path(const std::string& package_name, const std::string& message_name);
   MsgSpec parse_msg_file(const std::string& package_name, const std::string& file_path);
   std::shared_ptr<rospack::Rosstackage> rospack_;
-  std::map<std::string, MsgSpec> msg_spec_cache_;
+  std::unordered_map<std::string, MsgSpec> msg_spec_cache_;
 };
 
 }  // namespace foxglove_bridge
--- a/ros1_foxglove_bridge/include/foxglove_bridge/msg_parser.hpp
+++ b/ros1_foxglove_bridge/include/foxglove_bridge/msg_parser.hpp
@@ -1,0 +1,49 @@
+#pragma once
+
+#include <map>
+#include <memory>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+namespace rospack {
+class Rosstackage;
+}
+
+namespace foxglove_bridge {
+
+struct MsgSpecException : public std::runtime_error {
+  using std::runtime_error::runtime_error;
+};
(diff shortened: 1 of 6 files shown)


ros1_foxglove_bridge/include/foxglove_bridge/msg_parser.hpp
```diff
@@ -0,0 +1,49 @@
...
+  /// The output is equal to what `rosrun roslib gendeps -c <msg_file>` generates.
+  /// @param message_type Message type in the format <package_name>/<message_name>
+  /// @return String with the message schema and concatenated schemas of submessages.
+  std::string get_message_schema(const std::string& message_type);
+
+private:
+  std::string get_msg_file_path(const std::string& package_name, const std::string& message_name);
+  MsgSpec parse_msg_file(const std::string& package_name, const std::string& file_path);
+  std::shared_ptr<rospack::Rosstackage> rospack_;
+  std::map<std::string, MsgSpec> msg_spec_cache_;
```
jhurliman, line 46: Does this need to be sorted alphabetically or can it use `std::unordered_map`?

ros1_foxglove_bridge/src/msg_parser.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+      if (is_builtin_type(field_type)) {
+        continue;
+      } else if (is_header_type(field_type)) {
+        spec.types.push_back(HEADER);
+      } else if (field_type.find("/") == std::string::npos) {
+        // Message type specifier without package_name. References messages in same package.
+        spec.types.push_back(package_name + "/" + field_type);
+      } else {
+        // Explicity type specifier: <package_name>/<message_name>.
```
jhurliman, line 111: "Explicit"

jhurliman (approved):
Looks good. One step closer to ROS 1 bridge!

ros1_foxglove_bridge/include/foxglove_bridge/msg_parser.hpp
achim-k: You are right, an unordered map is better here

jhurliman:
```
/ros1_ws/src/ros-foxglove-bridge/ros1_foxglove_bridge/src/msg_parser.cpp:1:10: fatal error: filesystem: No such file or directory
 #include <filesystem>
          ^~~~~~~~~~~~
compilation terminated.
```

You can run `make melodic-test` locally while iterating on a fix for this. I believe the compiler and standard library shipping in Ubuntu 18.04 is too old to use the STL filesystem without the experimental prefix. Something like...

```cpp
#if __has_include(<filesystem>)
  #include <filesystem>
  namespace fs = std::filesystem;
#elif __has_include(<experimental/filesystem>)
  #include <experimental/filesystem> 
  namespace fs = std::experimental::filesystem;
#else
  error "Missing the <filesystem> header."
#endif
```

Taken from https://askubuntu.com/questions/1256440/how-to-get-libstdc-with-c17-filesystem-headers-on-ubuntu-18-bionic

achim-k:
Forgot to add install rules, just added them

PR #22, opened 2022-11-10
Make ros1 version nodelet compatible.

achim-k:
**Public-Facing Changes**
Adds the possibility to run the future ros1 server as a nodelet


achim-k:
It seems that rosdep dependencies are not installed :thinking: Will look into it.

jhurliman:
> It seems that rosdep dependencies are not installed thinking Will look into it.

You have to change the Dockerfiles to install the dependencies. We don't execute a "rosdep install" or whatever the command would be to tell ROS to run `apt-get install`.

achim-k:
> > It seems that rosdep dependencies are not installed thinking Will look into it.
> 
> You have to change the Dockerfiles to install the dependencies. We don't execute a "rosdep install" or whatever the command would be to tell ROS to run `apt-get install`. There are three relevant Dockerfiles:
> 
> * `./Dockerfile.ros1`
> * `./Dockerfile.rios1bionic`
> * `.devcontainer/Dockerfile`

#26. Didn't touch the `.devcontainer` Dockerfile though, the developer will have to run `rosdep install ...` every time they add new dependencies.

jhurliman:
> > > It seems that rosdep dependencies are not installed thinking Will look into it.
> > 
> > 
> > You have to change the Dockerfiles to install the dependencies. We don't execute a "rosdep install" or whatever the command would be to tell ROS to run `apt-get install`. There are three relevant Dockerfiles:
> > 
> > * `./Dockerfile.ros1`
> > * `./Dockerfile.rios1bionic`
> > * `.devcontainer/Dockerfile`
> 
> #26. Didn't touch the `.devcontainer` Dockerfile though, the developer will have to run `rosdep install ...` every time they add new dependencies.

I think it would be useful to automatically install both ROS 1 and ROS 2 dependencies as a step in the `.devcontainer/Dockerfile`. And remove all of the manual dependency installs from each of the Dockerfiles since we use rosdep to do that now.

jhurliman (approved):

achim-k:
> I think it would be useful to automatically install both ROS 1 and ROS 2 dependencies as a step in the `.devcontainer/Dockerfile`. And remove all of the manual dependency installs from each of the Dockerfiles since we use rosdep to do that now.

I think so too. We can probably also unify the Dockerfiles if we do so. The `rosdep` executable is part of the `ros:<distro>-base` image, so maybe we could switch to that, so we don't have to care about different python versions in melodic / noetic

PR #26, opened 2022-11-11
Add rosdep installation step to CI dockerfiles

achim-k:
**Public-Facing Changes**
None

**Description**
Adds rosdep init & rosdep install steps to the Dockerfiles which are used for CI. This will automatically install new dependencies that are added to `package.xml` without having to manually add them to the list of installed packages.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -19,6 +19,7 @@
     <depend condition="$ROS_VERSION == 1">rospack</depend>
     <depend condition="$ROS_VERSION == 2">ament_index_cpp</depend>
     <depend condition="$ROS_VERSION == 2">rclcpp</depend>
+    <depend>libwebsocketpp-dev</depend>
 
     <!-- test dependencies -->
     <build_depend condition="$ROS_VERSION == 1">rostest</build_depend>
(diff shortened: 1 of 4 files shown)


package.xml
```diff
@@ -19,6 +19,7 @@
     <depend condition="$ROS_VERSION == 1">rospack</depend>
     <depend condition="$ROS_VERSION == 2">ament_index_cpp</depend>
     <depend condition="$ROS_VERSION == 2">rclcpp</depend>
+    <depend>libwebsocketpp-dev</depend>
```
defunctzombie, line 22: What makes this a depend vs build_depend

achim-k: This is a legit question. I don't know why I changed it, it should have remained a build dependency since it is only required during build.

jhurliman: It's a `<build_depend>` because websocketpp is a header-only library. If it had a shared runtime library there would be a package called `websocketpp` that would be a `<depend>` while libwebsoscketpp-dev remained a `<build_depend>`

achim-k: -> #27

PR #29, opened 2022-11-11
Add ROS1 subscription support

achim-k:
**Public-Facing Changes**
Advertises ROS1 topics as channels and allows to subscribe to them e.g. from Foxglove Studio.

**Description**
Implementation is mainly based on https://github.com/achim-k/foxglove_websocket_server

One mayor problem that I noticed, is that the message schema parsing approach (parsing .msg files from the disk) does not work if topics are published from a bag file. We will have to think about what to do here, maybe the solution is then to fallback and subscribe to the topic once to get the message definition from the connection header.

### DIFF
--- a/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
@@ -21,7 +21,6 @@
 namespace foxglove {
 
 using json = nlohmann::json;
-using namespace std::placeholders;
 
 using ConnHandle = websocketpp::connection_hdl;
 using OpCode = websocketpp::frame::opcode::value;
@@ -166,11 +165,13 @@ inline Server<ServerConfiguration>::Server(std::string name, LogCallback logger)
 
   _server.clear_access_channels(websocketpp::log::alevel::all);
   _server.set_access_channels(APP);
-  _server.set_tcp_pre_init_handler(std::bind(&Server::socketInit, this, _1));
-  _server.set_validate_handler(std::bind(&Server::validateConnection, this, _1));
-  _server.set_open_handler(std::bind(&Server::handleConnectionOpened, this, _1));
-  _server.set_close_handler(std::bind(&Server::handleConnectionClosed, this, _1));
-  _server.set_message_handler(std::bind(&Server::handleMessage, this, _1, _2));
+  _server.set_tcp_pre_init_handler(std::bind(&Server::socketInit, this, std::placeholders::_1));
+  _server.set_validate_handler(std::bind(&Server::validateConnection, this, std::placeholders::_1));
+  _server.set_open_handler(std::bind(&Server::handleConnectionOpened, this, std::placeholders::_1));
+  _server.set_close_handler(
+    std::bind(&Server::handleConnectionClosed, this, std::placeholders::_1));
+  _server.set_message_handler(
+    std::bind(&Server::handleMessage, this, std::placeholders::_1, std::placeholders::_2));
   _server.set_reuse_addr(true);
   _server.set_listen_backlog(128);
 }
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,13 +75,13 @@ if("$ENV{ROS_VERSION}" STREQUAL "1")
     message(STATUS "Building with catkin")
     set(ROS_BUILD_TYPE "catkin")
 
-    find_package(catkin REQUIRED COMPONENTS nodelet roslib roscpp rospack)
+    find_package(catkin REQUIRED COMPONENTS nodelet ros_type_introspection roslib roscpp rospack)
     find_package(Boost REQUIRED)
 
     catkin_package(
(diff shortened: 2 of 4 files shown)


defunctzombie:
> One mayor problem that I noticed, is that the message schema parsing approach (parsing .msg files from the disk) does not work if topics are published from a bag file. We will have to think about what to do here, maybe the solution is then to fallback and subscribe to the topic once to get the message definition from the connection header.

What does rosbridge do? It too doesn't have the schemas on disk when playing from a bag file yet I believe we are able to receive the entire topics and datatypes into Studio?

jhurliman:
`rosbridge` (`ros1`) branch does something like this:

```
roslib.launcher.load_manifest(modname)
pypkg = __import__('%s.%s' % (modname, subname))
cls = getattr(getattr(pypkg, subname), classname)
instance = cls()
return instance._full_text
```

Which suggests it is loading genpy-generated Python code from disk based on the datatype and extracting the message definition string from the generated Python.

jhurliman:
I wonder if it may be better to unconditionally use the socket-based approach. It has the overhead of opening a socket, initiating a two-way handshake, then closing the socket, only to reopen it again if the client requests a subscription. But even for 1000-topic graphs this is a one-time cost of opening and then closing 1000 sockets. I don't think that's a bad price to pay for always having the correct message definitions and not worrying whether what is locally on disk matches what is actually being advertised.

achim-k:
I opened #33 to track this problem. I don't think this is something we have to fix immediately as also `rosbridge` suffers from the same problem. It is also not very common to play bag bagfiles for which no local message definitions are available.

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -166,11 +166,13 @@ inline Server<ServerConfiguration>::Server(std::string name, LogCallback logger)
...
-  _server.set_open_handler(std::bind(&Server::handleConnectionOpened, this, _1));
-  _server.set_close_handler(std::bind(&Server::handleConnectionClosed, this, _1));
-  _server.set_message_handler(std::bind(&Server::handleMessage, this, _1, _2));
+  _server.set_tcp_pre_init_handler(std::bind(&Server::socketInit, this, std::placeholders::_1));
+  _server.set_validate_handler(std::bind(&Server::validateConnection, this, std::placeholders::_1));
+  _server.set_open_handler(std::bind(&Server::handleConnectionOpened, this, std::placeholders::_1));
+  _server.set_close_handler(
+    std::bind(&Server::handleConnectionClosed, this, std::placeholders::_1));
+  _server.set_message_handler(
+    std::bind(&Server::handleMessage, this, std::placeholders::_1, std::placeholders::_2));
```
jhurliman, line 175: It's easier to read these using the `_1` `_2` shorthand, since we're using the std::placeholders namespace above.

ros1_foxglove_bridge/include/foxglove_bridge/subscription.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include <ros/subscriber.h>
+
+namespace foxglove_bridge {
+
+class Subscription {
+public:
+  explicit Subscription(ros::NodeHandle& nh, const std::string& topic)
+      : nh_(nh)
+      , topic_(topic)
+      , num_subscribed_clients_(0)
```
jhurliman, line 19: websocket_server.hpp already does subscription count tracking and only fires the subscribe/unsubscribe callbacks when subscription count goes from 0->1 or 1->0.

jhurliman (changes requested):
The `Subscription` class is doing redundant work with it's `num_subscribed_clients_` tracking. Once that is removed the rest LGTM

ros1_foxglove_bridge/include/foxglove_bridge/subscription.hpp
achim-k: Ah ok great, I have missed this!

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
achim-k: I had to use the full namespace specifier here as it would otherwise collide with boost placeholders. We include  (transitively) boost headers as ROS1 depends on it.

achim-k: I think we have the following options:
1. Remove the `using std::placeholders` statement and use the explicit namespace 
2. Move the server definition to a cpp file (would require explicit template instantiation for all `ServerConfiguration`s we want to support)

jhurliman: Let's remove `using std::placeholders`. This is a header-only implementation since it will be merged into the upstream `ws-protocol` repo and act as a reference server implementation that can be copy-pasted into other C++ projects as a single file.

jhurliman (approved):

PR #32, opened 2022-11-11
Switch foxglove_bridge_base back to a STATIC library

jhurliman:
**Public-Facing Changes**

**Description**
It wasn't clear why this was switched from a STATIC to SHARED library, but I think it broke the ROS 2 node.


achim-k:
In https://github.com/foxglove/ros-foxglove-bridge/pull/34 I added the missing install rule which makes the ros2 node work again

PR #39, opened 2022-11-11
Dynamically enable/disable simulated time based on /clock topic presence

jhurliman:
**Public-Facing Changes**
- Message timestamps are set to sim time if `/clock` is published, `use_sim_time` parameter is no longer required

**Description**
This simplifies execution of the bridge by dynamically toggling between sim time and system time based on the presence of `/clock`.


jhurliman:
With this change, playing back the ROS 2 nuScenes MCAP from the benchmark assets seems visually correct and stable (except for the CameraInfo messages which are corrupt, need to regenerate the file with an additional fix).

PR #50, opened 2022-11-16
Add address parameter

### LINKED ISSUE #36: Support "address" parameter
Control which interface we bind to using an `"address"` parameter

achim-k:
**Public-Facing Changes**
Add `address` parameter

**Description**

Adds the `address` parameter which is forwarded to [websocketpp::transport::asio::endpoint<config>::listen](https://docs.websocketpp.org/classwebsocketpp_1_1transport_1_1asio_1_1endpoint.html#a85606665cc9c948194076c2377cb61c0) `host` parameter:

> host | A string identifying a location. May be a descriptive name or a numeric address string.
> -- | --

Fixes #36

### DIFF
--- a/ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
+++ b/ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
@@ -55,6 +55,8 @@ public:
                                                    &FoxgloveBridge::updateAdvertisedTopics, this);
     } catch (const std::exception& err) {
       ROS_ERROR("Failed to start websocket server: %s", err.what());
+      // Rethrow exception such that the nodelet is unloaded.
+      throw err;
     }
   };
   virtual ~FoxgloveBridge() {
--- a/ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
+++ b/ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
@@ -19,7 +19,7 @@
 namespace foxglove_bridge {
 
 constexpr int DEFAULT_PORT = 8765;
-constexpr char DEFAULT_HOST[] = "0.0.0.0";
+constexpr char DEFAULT_ADDRESS[] = "0.0.0.0";
 constexpr int DEFAULT_MAX_UPDATE_MS = 5000;
 constexpr char ROS1_CHANNEL_ENCODING[] = "ros1";
 constexpr uint32_t SUBSCRIPTION_QUEUE_LENGTH = 10;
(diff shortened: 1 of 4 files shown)


jhurliman:
Is there a way to add a parameter alias? I'm wondering if we can ease the transition for people coming from rosbridge, since we use identical parameter names except for this one.

achim-k:
> Is there a way to add a parameter alias? I'm wondering if we can ease the transition for people coming from rosbridge, since we use identical parameter names except for this one.

Then I would say we just call the parameter `address` instead?

jhurliman:
I expect you would typically put in an IP address to bind to. Would using a DNS hostname even work?

ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
```diff
@@ -32,24 +33,29 @@ class FoxgloveBridge : public nodelet::Nodelet {
...
+        std::bind(&FoxgloveBridge::subscribeHandler, this, std::placeholders::_1));
+      _server->setUnsubscribeHandler(
+        std::bind(&FoxgloveBridge::unsubscribeHandler, this, std::placeholders::_1));
+      _server->start(host, static_cast<uint16_t>(port));
+
+      _msgParser = std::make_unique<foxglove_bridge::MsgParser>();
+      _updateTimer = getMTNodeHandle().createTimer(ros::Duration(max_update_ms / 1e3),
+                                                   &FoxgloveBridge::updateAdvertisedTopics, this);
+    } catch (const std::exception& err) {
+      ROS_ERROR("Failed to start websocket server: %s", err.what());
```
jhurliman, line 57: This is not a good state to leave the node in. There is no retry loop or ability to try again, so if the server failed to start it would print to a log somewhere and just be a zombie node at that point. Better to crash the node.

achim-k: It's a bit more involved as the server is loaded as a nodelet. So if you would try to exit the process then the nodelet manager and all other loaded nodelets would terminate as well. Will check what the common way to handle this is.

achim-k:
> I expect you would typically put in an IP address to bind to. Would using a DNS hostname even work?

It works e.g. with `localhost` or whatever is set as `hostname`

ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
jhurliman: I noticed rosbridge has some somewhat complicated retry loop code. Maybe we can come up with something simpler, but that's always a last resort option.

achim-k: Nodelets that fail to initialize are unloaded again:

https://github.com/ros/nodelet_core/blob/6265cac6bd686a7feb872bba5304f16de4f0fd46/nodelet/src/loader.cpp#L314-L326

The referenced exception handler doesn't log the specific error message, so I would catch it to log an error and then retrow it.

jhurliman (approved):

PR #53, opened 2022-11-17
Add support for latched topics (or topics published with transient local policy)

### LINKED ISSUE #30: ROS 1 latched topics
To make topics like `/tf_static` work, latching semantics must be carried through the bridge. Upon subscribing to a latched topic, the most recent message must be held in memory and sent when a client subscribes to the latched channel.

### LINKED ISSUE #31: ROS 2 "latched" topics
To make topics like `/tf_static` work, transient-local semantics must be carried through the bridge. Upon subscribing to a topic with one or more transient-local publishers, transient-local should be enabled, and the depth should equal the max depth of the set of transient-local publishers. An in-memory buffer of size `depth` should hold the latest messages, and all of those messages sent to new subscribers.

achim-k:
**Public-Facing Changes**
Add support for topics that are latched (ROS1) or published with transient local durability policy (ROS2)

**Description**
Instead of having only a single subscriber instance per topic, we now create a subscriber for every client that subscribes to the topic. This way we do not have to re-implement the latching logic and we instead rely on ROS to correctly send latched messages to new subscriber instances.

Update:
- Mirrors subscription qos profile selection of `ros2 topic echo` ([source](https://github.com/ros2/ros2cli/blob/619b3d1c9b38f49bfd2a9286b1086d8eed1ecb94/ros2topic/ros2topic/verb/echo.py#L137-L194))
- Allows setting a max QoS depth via ROS2 parameter

Fixes #30 
Fixes #31

### DIFF
--- a/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
@@ -567,19 +567,23 @@ inline void Server<ServerConfiguration>::sendMessage(ChannelId chanId, uint64_t 
 template <typename ServerConfiguration>
 inline void Server<ServerConfiguration>::sendMessage(ConnHandle clientHandle, ChannelId chanId,
                                                      uint64_t timestamp, std::string_view data) {
-  std::shared_lock<std::shared_mutex> lock(_clientsChannelMutex);
+  SubscriptionId subId = std::numeric_limits<SubscriptionId>::max();
 
-  const auto clientHandleAndInfoIt = _clients.find(clientHandle);
-  if (clientHandleAndInfoIt == _clients.end()) {
-    return;  // Client got removed in the meantime.
+  {
+    std::shared_lock<std::shared_mutex> lock(_clientsChannelMutex);
+    const auto clientHandleAndInfoIt = _clients.find(clientHandle);
+    if (clientHandleAndInfoIt == _clients.end()) {
+      return;  // Client got removed in the meantime.
+    }
+
+    const auto& client = clientHandleAndInfoIt->second;
+    const auto& subs = client.subscriptionsByChannel.find(chanId);
+    if (subs == client.subscriptionsByChannel.end()) {
+      return;  // Client not subscribed to this channel.
+    }
+    subId = subs->second;
   }
 
-  const auto& client = clientHandleAndInfoIt->second;
-  const auto& subs = client.subscriptionsByChannel.find(chanId);
-  if (subs == client.subscriptionsByChannel.end()) {
-    return;  // Client not subscribed to this channel.
-  }
-  const auto subId = subs->second;
   std::vector<uint8_t> message(1 + 4 + 8 + data.size());
   message[0] = uint8_t(BinaryOpcode::MESSAGE_DATA);
   foxglove::WriteUint32LE(message.data() + 1, subId);
--- a/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
@@ -82,6 +82,7 @@ public:
   using ConnectionType = websocketpp::connection<ServerConfiguration>;
   using MessagePtr = typename ServerType::message_ptr;
   using Tcp = websocketpp::lib::asio::ip::tcp;
+  using SubscribeUnsubscribeHandler = std::function<void(ChannelId, ConnHandle)>;
 
   static const std::string SUPPORTED_SUBPROTOCOL;
 
@@ -100,10 +101,12 @@ public:
   void removeChannel(ChannelId chanId);
   void broadcastChannels();
 
-  void setSubscribeHandler(std::function<void(ChannelId)> handler);
-  void setUnsubscribeHandler(std::function<void(ChannelId)> handler);
+  void setSubscribeHandler(SubscribeUnsubscribeHandler handler);
+  void setUnsubscribeHandler(SubscribeUnsubscribeHandler handler);
 
   void sendMessage(ChannelId chanId, uint64_t timestamp, std::string_view data);
+  void sendMessage(ConnHandle clientHandle, ChannelId chanId, uint64_t timestamp,
+                   std::string_view data);
 
   std::optional<Tcp::endpoint> localEndpoint();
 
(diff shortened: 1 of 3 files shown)


ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -324,35 +330,50 @@ class FoxgloveBridge : public rclcpp::Node {
...
+      qos.transient_local();
+    }
 
     try {
       auto subscriber = this->create_generic_subscription(
         topic, datatype, qos,
-        [&](std::shared_ptr<rclcpp::SerializedMessage> msg) {
-          rosMessageHandler(channel, msg);
-        },
+        std::bind(&FoxgloveBridge::rosMessageHandler, this, channel, clientHandle, _1),
```
achim-k, line 357: Had to convert this from a lambda function to std::bind to avoid segmentation faults. It would have probably also worked with a lambda that captures `=clientHandle` (as copy).

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -567,6 +564,30 @@ inline void Server<ServerConfiguration>::sendMessage(ChannelId chanId, uint64_t
...
+  if (subs == client.subscriptionsByChannel.end()) {
+    return;  // Client not subscribed to this channel.
+  }
+  const auto subId = subs->second;
+  std::vector<uint8_t> message(1 + 4 + 8 + data.size());
+  message[0] = uint8_t(BinaryOpcode::MESSAGE_DATA);
+  foxglove::WriteUint32LE(message.data() + 1, subId);
+  foxglove::WriteUint64LE(message.data() + 5, timestamp);
+  std::memcpy(message.data() + 1 + 4 + 8, data.data(), data.size());
+  sendBinary(clientHandle, message);
```
jhurliman, line 588: I think it would be better to release the mutex before calling `sendBinary()`

ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
```diff
@@ -26,7 +26,7 @@ constexpr uint32_t SUBSCRIPTION_QUEUE_LENGTH = 10;
 
 using ServerType = foxglove::Server<foxglove::WebSocketNoTls>;
 using TopicAndDatatype = std::pair<std::string, std::string>;
-using Subscription = ros::Subscriber;
+using SubscriptionsByClient = std::map<foxglove::ConnHandle, ros::Subscriber, std::owner_less<>>;
```
jhurliman, line 29: `std::unordered_map`, unless `std::map` is required in order to use `std::owner_less`?

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -324,35 +330,50 @@ class FoxgloveBridge : public rclcpp::Node {
...
-      if (publisher.qos_profile().reliability() == rclcpp::ReliabilityPolicy::BestEffort) {
+      const auto& qos = publisher.qos_profile();
+      if (qos.reliability() == rclcpp::ReliabilityPolicy::BestEffort) {
         reliable = false;
-        break;
       }
+      if (qos.durability() == rclcpp::DurabilityPolicy::TransientLocal) {
+        durationVolatile = false;
+      }
+      depth += qos.depth();
```
jhurliman, line 343:
```suggestion
      constexpr size_t MAX_QUEUE_DEPTH = 10;
      depth = std::min(MAX_QUEUE_DEPTH, std::max(depth, qos.depth()));
```

jhurliman (approved):
Couple of suggestions, otherwise LGTM

ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
achim-k: Yes it only works with `std::map`. I took it from https://github.com/foxglove/ros-foxglove-bridge/blob/b3faff643e5a8aec1c2592896a3609f1b4165384/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp#L129

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
achim-k: I tested it locally with having 4 transient local publishers and it only works if the depth is at least as big as the amount of publishers.

```bash
# Launch 4 publishers
for i in {1..4}; do ros2 topic pub --qos-history keep_last --qos-depth 1 --qos-durability transient_local -1 --keep-alive 100 /foo std_msgs/String 'data: "hi"' & done
```

```bash
# Subscribe with depth = 1
$ ros2 topic echo --qos-durability transient_local --qos-depth 1 --qos-reliability reliable /foo
data: hi
---
```



```bash
# subscribe with default depth (not sure what the default is)
$ ros2 topic echo --qos-durability transient_local --qos-reliability reliable /foo
data: hi
---
data: hi
---
data: hi
---
data: hi
---
```

achim-k: Maybe we can make it the following?

```cpp
constexpr size_t MAX_QUEUE_DEPTH = 10;
depth = std::min(MAX_QUEUE_DEPTH, depth + qos.depth());

achim-k: Maybe the best we could do is to replicate how `ros2 topic echo` chooses the subscriber qos?

https://github.com/ros2/ros2cli/blob/619b3d1c9b38f49bfd2a9286b1086d8eed1ecb94/ros2topic/ros2topic/verb/echo.py#L137-L194

achim-k: `ros2 topic echo` uses the [qos_profile_sensor_data](https://github.com/ros2/rmw/blob/31c6fd094c8bd01d0a231856df1bd9a476bea26a/rmw/include/rmw/qos_profiles.h#L25-L36) qos preset which has a default depth of 5

achim-k: It probably makes sense to add a parameter that allows to choose one of the [qos presets](https://github.com/ros2/rmw/blob/31c6fd094c8bd01d0a231856df1bd9a476bea26a/rmw/include/rmw/qos_profiles.h) as the default subscription qos profile

jhurliman (approved):

PR #57, opened 2022-11-18
Add SSL server support

### LINKED ISSUE #14: TLS support
The current WebSocket server is a template instantiation using the "NoTls" configuration. We need to support TLS (`wss://`) if the relevant parameters are set at startup.

achim-k:
**Public-Facing Changes**
Allows to start the server with SSL enabled

**Description**
This PR adds TLS support which can be enabled by setting the corresponding `tls` parameter.

Excerpt from README:

>  * __tls__: If `true`, use Transport Layer Security (TLS) for encrypted communication. Defaults to `false`.
>  * __certfile__: Path to the certificate to use for TLS. Required when __tls__ is set to `true`. Defaults to `""`.
>  * __keyfile__: Path to the private key to use for TLS. Required when __tls__ is set to `true`. Defaults to `""`.

I could not get this working with a self signed certificate. Only made it work with a certificate generated with let's encrypt (added an entry in `/etc/hosts` to resolve the name to my local docker container)

Fixes #14

### DIFF
--- a/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
@@ -17,6 +17,7 @@
 #include "serialization.hpp"
 #include "websocket_logging.hpp"
 #include "websocket_notls.hpp"
+#include "websocket_tls.hpp"
 
 namespace foxglove {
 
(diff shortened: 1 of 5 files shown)


foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -602,4 +639,51 @@ inline std::optional<asio::ip::tcp::endpoint> Server<ServerConfiguration>::local
...
+      constexpr char ciphers[] =
+        "ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:"
+        "ECDHE-ECDSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-DSS-AES128-GCM-SHA256:kEDH+"
+        "AESGCM:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-"
+        "AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA:ECDHE-"
+        "ECDSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-DSS-AES128-SHA256:DHE-RSA-"
+        "AES256-SHA256:DHE-DSS-AES256-SHA:DHE-RSA-AES256-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:"
+        "AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:AES:CAMELLIA:DES-CBC3-SHA:!aNULL:!eNULL:"
+        "!EXPORT:!DES:!RC4:!MD5:!PSK:!aECDH:!EDH-DSS-DES-CBC3-SHA:!EDH-RSA-DES-CBC3-SHA:!KRB5-DES-"
+        "CBC3-SHA";
```
jhurliman, line 676: Can we provide a link or reference to where this cipher list came from?

jhurliman (approved):

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
achim-k: It stems from the websocketpp example echo tls server:
https://github.com/zaphoyd/websocketpp/blob/1b11fd301531e6df35a6107c1e8665b1e77a2d8e/examples/echo_server_tls/echo_server_tls.cpp#L121

But I actually should have used the more restrictive version 2 lines above since we also disable `tlsv1`. I will update the cipher and also add a link to the example server code.

Issue #58, opened 2022-11-18
ROS 2 message definition search assumes `msg/` in the path, doesn't handle subfolders of message definitions


jhurliman:
`message_definition_cache.cpp` should use `ament_index_cpp::get_resources("rosidl_interfaces")` to get a list of package names, then call `ament_index_cpp::get_resource("rosidl_interfaces", packageName, relativePathList, basePath)` to get relativePathList that can be split on newline, iterated, filtered for .msg suffix and matched against a target datatype, then combined with basePath to form an absolute path to the target msgdef.

PR #61, opened 2022-11-21
Handle other exceptions when getting ros2 message definitions

### LINKED ISSUE #60: Bridge not compatible with ROS2 actions
When launching the foxglove bridge in ROS2, it will crash if an action server is present.

- Version: latest
- Platform: ubuntu jammy, ros2 humble

**Steps To Reproduce**
1. Launch the getting started example for Nav2: https://navigation.ros.org/getting_started/index.html#running-the-example
Specifically, the following:
```
export TURTLEBOT3_MODEL=waffle
export GAZEBO_MODEL_PATH=$GAZEBO_MODEL_PATH:/opt/ros/<ros2-distro>/share/turtlebot3_gazebo/models
ros2 launch nav2_bringup tb3_simulation_launch.py headless:=False
```
Alternately, you can probably reproduce by launching anything that has an action server or client. 

2. Run the foxglove bridge:
```
ros2 run foxglove_bridge foxglove_bridge
```
The bridge will exit with an error (described below). 

**Expected Behavior**
Bridge launches.

**Actual Behavior**
Bridge crashes with error:
```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Invalid package resource name: nav2_msgs/action/NavigateThroughPoses_FeedbackMessage
```

achim-k:
**Public-Facing Changes**
Improve error handling when retrieving message definitions (ros2)

**Description**
The presence of action topics causes the bridge to crash:

```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Invalid package resource name: turtlesim/action/RotateAbsolute_FeedbackMessage
```

The reason why the exception is thrown, is because we assume that message files are located in the `msg` folder:
https://github.com/foxglove/ros-foxglove-bridge/blob/08ff9e522065a2c9998b0fbf2030d9e8fa363637/ros2_foxglove_bridge/src/message_definition_cache.cpp#L19

This PR for now only adds an additional exception handler that will simply log a warning. Proper action support will be implemented later on.

Fixes #60

### DIFF
--- a/ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
+++ b/ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
@@ -236,6 +236,10 @@ public:
                       topicAndDatatype.second.c_str(), err.what());
           // We still advertise the channel, but with an emtpy schema
           newChannel.schema = "";
+        } catch (const std::exception& err) {
+          RCLCPP_WARN(this->get_logger(), "Failed to add channel for topic \"%s\" (%s): %s",
+                      topicAndDatatype.first.c_str(), topicAndDatatype.second.c_str(), err.what());
+          continue;
         }
 
         auto channel = foxglove::Channel{_server.addChannel(newChannel), newChannel};


jhurliman (approved):

jhurliman:
I think implementing https://github.com/foxglove/ros-foxglove-bridge/issues/58 will get us closer to a proper fix for this as well.

PR #64, opened 2022-11-22
Use ros_babel_fish for getting ROS1 message definitions

achim-k:
**Public-Facing Changes**
Use ros_babel_fish for getting ROS1 message definitions

**Description**
Make use of [ros_babel_fish](https://github.com/StefanFabian/ros_babel_fish) instead of maintaining our own cpp library to parse ROS1 message definitions. This will also allow us to retrieve the md5sum later on


jhurliman (approved):

jhurliman:
Have you checked that this returns a concatenated message definition (with ======… separators)?

achim-k:
> Have you checked that this returns a concatenated message definition (with ======… separators)?

Yes, it works correctly. For reference, this is the output for `std_msgs/String`:

```
##
## Severity level constants
##
byte DEBUG=1 #debug level
byte INFO=2  #general level
byte WARN=4  #warning level
byte ERROR=8 #error level
byte FATAL=16 #fatal/critical level
##
## Fields
##
Header header
byte level
string name # name of the node
string msg # message 
string file # file the message came from
string function # function the message came from
uint32 line # line the message came from
string[] topics # topic names that the node publishes

================================================================================
MSG: std_msgs/Header
# Standard metadata for higher-level stamped data types.
# This is generally used to communicate timestamped data 
# in a particular coordinate frame.
# 
# sequence ID: consecutively increasing ID 
uint32 seq
#Two-integer timestamp that is expressed as:
# * stamp.sec: seconds (stamp_secs) since epoch (in Python the variable is called 'secs')
# * stamp.nsec: nanoseconds since stamp_secs (in Python the variable is called 'nsecs')
# time-handling sugar is provided by the client library
time stamp
#Frame this data is associated with
string frame_id
```

PR #68, opened 2022-11-23
Fix ros1 topic discovery function not being called frequently at startup

achim-k:
**Public-Facing Changes**
Fix ros1 topic discovery function not being called frequently at startup

**Description**
Topic discovery was called for the first time at `max_update_ms` (defaults to 5s) after startup. This patch fixes this and also adds a minimum duration of 100ms between two consecutive topic discoveries

Edit: Also fixes the `_maxUpdateTimer` member never being set from the ROS param

### DIFF
--- a/ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
+++ b/ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
@@ -23,6 +23,7 @@ constexpr char DEFAULT_ADDRESS[] = "0.0.0.0";
 constexpr int DEFAULT_MAX_UPDATE_MS = 5000;
 constexpr char ROS1_CHANNEL_ENCODING[] = "ros1";
 constexpr uint32_t SUBSCRIPTION_QUEUE_LENGTH = 10;
+constexpr double MIN_UPDATE_PERIOD_MS = 100.0;
 
 using TopicAndDatatype = std::pair<std::string, std::string>;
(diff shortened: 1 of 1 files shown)


ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
```diff
@@ -59,7 +60,7 @@ class FoxgloveBridge : public nodelet::Nodelet {
                                                std::placeholders::_1, std::placeholders::_2));
       _server->start(address, static_cast<uint16_t>(port));
 
-      _updateTimer = getMTNodeHandle().createTimer(ros::Duration(max_update_ms / 1e3),
+      _updateTimer = getMTNodeHandle().createTimer(ros::Duration(MIN_UPDATE_PERIOD_MS / 1e3),
```
jhurliman, line 63: Why create a timer here at all? I think you can just directly call `updateAdvertisedTopics()`

ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
```diff
@@ -269,7 +270,8 @@ class FoxgloveBridge : public nodelet::Nodelet {
 
     // Schedule the next update using truncated exponential backoff, up to `_maxUpdateMs`
```
jhurliman, line 271:
```suggestion
    // Schedule the next update using truncated exponential backoff, between `MIN_UPDATE_PERIOD_MS` and `_maxUpdateMs`
```

jhurliman (approved):

jhurliman (approved):

PR #77, opened 2022-11-29
Add ros2 smoke test

### LINKED ISSUE #20: Smoke testing
rosbridge has CI tests that start a rosbridge node, connect to it with a Python WebSocket client, and confirm the connection can be opened and messages exchanged before tearing it down. It would be nice to have this level of end-to-end testing for our ROS 1 and ROS 2 nodes, although I'm not sure what the implementation would look like.

achim-k:
**Public-Facing Changes**
None

**Description**
Adds a smoke test for the ROS2 implementation. This also adds a no-op timer in the ROS 2 node with a 100ms frequency, this appears to be required to spin the node when repeatedly subscribing and unsubscribing to a topic.

Fixes #20

### DIFF
--- a/ros2_foxglove_bridge/tests/smoke_test.cpp
+++ b/ros2_foxglove_bridge/tests/smoke_test.cpp
@@ -54,10 +54,6 @@ TEST(SmokeTest, testSubscription) {
       }
     });
 
-    // Connect the client and wait for the channel future
-    ASSERT_EQ(std::future_status::ready, wsClient.connect(URI).wait_for(std::chrono::seconds(5)));
-    ASSERT_EQ(std::future_status::ready, channelFuture.wait_for(std::chrono::seconds(5)));
-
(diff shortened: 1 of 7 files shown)


ros2_foxglove_bridge/tests/smoke_test.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include <gtest/gtest.h>
+#include <rclcpp_components/component_manager.hpp>
+#include <std_msgs/msg/string.hpp>
+#include <websocketpp/config/asio_client.hpp>
+
+#include <foxglove_bridge/websocket_client.hpp>
+
+constexpr char URI[] = "ws://localhost:8765";
+
+// Binary representation of std_msgs/String for "hello world"
```
jhurliman, line 16:
```suggestion
// Binary representation of std_msgs/msg/String for "hello world"
```

PR #86, opened 2022-11-29
Revert waiting for graph change events and use timer approach again

achim-k:
This PR targets #77 

**Public-Facing Changes**
Fixes clients not receiving messages when a topic is resubscribed

**Description**
Effectively reverts 36012140a9a13a6ed82aeb9d69969fc8bc756c78 which caused issues when resubscribing to a topic

### DIFF
--- a/ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
+++ b/ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
@@ -15,6 +15,7 @@
 
 constexpr uint16_t DEFAULT_PORT = 8765;
 constexpr char DEFAULT_ADDRESS[] = "0.0.0.0";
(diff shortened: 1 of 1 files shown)


jhurliman:
We should get to the bottom of why this works first. I think it works not because the polling thread is broken, but because the periodic timer is causing something to happen in ROS 2 internals that kicks things alive. Instead of reverting the polling thread, if you add:

```cpp
_updateTimer = this->create_wall_timer(1s, std::bind(&FoxgloveBridge::doNothing, this));
```

It will also work. The timer itself is what "fixes" things.

jhurliman:
Closing, the issue is fixed in https://github.com/foxglove/ros-foxglove-bridge/pull/77/commits/99943a6b60d761a69276ce51a3131d1e550c5590

It would be worth creating a minimal repro to understand better what is going on.

PR #91, opened 2022-11-30
Force compile-time evaluation of opcode hashes

jhurliman:
**Public-Facing Changes**
- None

**Description**
Switched hashing algorithm to loop-based instead of recursion, force compile-time evaluation against constant opcode strings


foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -31,8 +31,12 @@ using OpCode = websocketpp::frame::opcode::value;
 static const websocketpp::log::level APP = websocketpp::log::alevel::app;
 static const websocketpp::log::level RECOVERABLE = websocketpp::log::elevel::rerror;
 
-constexpr uint32_t Integer(const char* str, uint32_t h = 0) {
-  return !str[h] ? 5381 : (Integer(str, h + 1) * 33) ^ str[h];
+constexpr uint32_t Integer(const std::string_view& str) {
```
jtbandes, line 34: I think it's generally advised to pass string_views by value instead of by reference

PR #95, opened 2022-11-30
Get message definition paths from the "rosidl_interfaces" ament index

### LINKED ISSUE #58: ROS 2 message definition search assumes `msg/` in the path, doesn't handle subfolders of message definitions

jhurliman:
**Public-Facing Changes**
- Fixes message definition loading for `.msg` or `.idl` files not located in `msg/`

**Description**
Previously, we hardcoded `/msg/` to build the absolute path to a message definition. This change uses the "rosidl_interfaces" ament index for a package to find the relative path to the target filename and construct an absolute path from that. Fixes #58

### DIFF
--- a/ros2_foxglove_bridge/src/message_definition_cache.cpp
+++ b/ros2_foxglove_bridge/src/message_definition_cache.cpp
@@ -15,6 +15,12 @@
 
 namespace foxglove {
 
+#ifdef _WIN32
+static const std::string SEPARATOR = "\\";
+#else
+static const std::string SEPARATOR = "/";
+#endif
+
 // Match datatype names (foo_msgs/Bar or foo_msgs/msg/Bar)
 static const std::regex PACKAGE_TYPENAME_REGEX{R"(^([a-zA-Z0-9_]+)/(?:msg/)?([a-zA-Z0-9_]+)$)"};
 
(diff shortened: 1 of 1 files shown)


ros2_foxglove_bridge/src/message_definition_cache.cpp
```diff
@@ -15,6 +15,12 @@
 
 namespace foxglove {
 
+#ifdef _WIN32
+static const std::string SEPARATOR = "\\";
+#else
+static const std::string SEPARATOR = "/";
+#endif
+
```
achim-k, line 23: alternative: [std::filesystem::path::preferred_separator](https://en.cppreference.com/w/cpp/filesystem/path)

achim-k (approved):

ros2_foxglove_bridge/src/message_definition_cache.cpp
jhurliman: I had that originally, but all of the #ifdefs required to include either experimental/filesystem or filesystem are a pain just to get this single character.

ros2_foxglove_bridge/src/message_definition_cache.cpp
```diff
@@ -117,15 +140,36 @@ const MessageSpec& MessageDefinitionCache::load_message_spec(
...
+  if (!ament_index_cpp::get_resource("rosidl_interfaces", package, index_contents)) {
+    throw DefinitionNotFoundError(definition_identifier.package_resource_name);
+  }
+
+  // Find the first line that ends with the filename we're looking for
+  const auto lines = split_string(index_contents);
+  const auto it = std::find_if(lines.begin(), lines.end(), [&filename](const std::string& line) {
+    return line.size() >= filename.size() &&
+           line.compare(line.size() - filename.size(), filename.size(), filename) == 0;
+  });
```
jtbandes, line 160: Does this work if a package contains both `AAAAFoo.msg` and `Foo.msg`, and we're searching for `Foo.msg`?

jhurliman: No, it will fail in that case 😭

PR #100, opened 2022-12-01
[ros2]: Improve finding of message file

### LINKED ISSUE #97: ROS 2: Handle `AAAAFoo.msg` and `Foo.msg` in the same package
_Originally posted by @jtbandes in https://github.com/foxglove/ros-foxglove-bridge/pull/95#discussion_r1036451404_

achim-k:
**Public-Facing Changes**
None

**Description**
Can now handle cases where a message file's name is a substring of another message file's name, e.g. `Foo.msg` and `AAAAFoo.msg`

Fixes #97

### DIFF
--- a/ros2_foxglove_bridge/src/message_definition_cache.cpp
+++ b/ros2_foxglove_bridge/src/message_definition_cache.cpp
@@ -16,12 +16,6 @@
 
 namespace foxglove {
 
-#ifdef _WIN32
-static const std::string SEPARATOR = "\\";
-#else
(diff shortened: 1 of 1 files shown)


jhurliman (approved):

jtbandes:
Is there a way we can add tests for these things?

achim-k:
> Is there a way we can add tests for these things?

We can, but we probably have to add an `foxglove_bridge_test_msgs` package for that which also means restructuring the repository.

PR #104, opened 2022-12-02
Stop spamming our contact form with buildfarm notifications

amacneil:


jtbandes (changes requested):
I think we should probably use the ros-tooling@foxglove.dev group which we made for this purpose. Whether John & Achim want to be listed individually as author/maintainer is up to them; personally I've liked to maintain [plausible deniability](https://github.com/foxglove/schemas/blob/main/ros_foxglove_msgs/package.xml#L13) for my own involvement in our published packages 😄

achim-k:
I believe that emails are only sent to maintainers, not authors. So #70 should have fixed this problem. Old release builds might still send emails to contact@foxglove.com but those should fade out soon. 

> I think we should probably use the ros-tooling@foxglove.dev group which we made for this purpose

That's a good idea. Maybe we could make it even ros-tooling+foxglove_bridge@foxglove.dev, so mail traffic for that project can be forwarded easily to the corresponding developers

amacneil:
I don't mind where they get sent, but contact@foxglove.dev should not be hooked up to any automated notifications (it goes to our sales inbox). We were still getting like 20 emails/day as of Thursday so I don't think https://github.com/foxglove/ros-foxglove-bridge/pull/70 worked.

jtbandes (approved):

PR #110, opened 2022-12-12
Optimize Dockerfiles

ruffsl:
I was looking at the source Dockerfile for the `ghcr.io/foxglove/rolling-ros2-bridge` image, and noticed some optimizations that could be made. See commit history for details.

The only major change is swapping the base image for the ros2 bridge from `ros-core` over to `ros-base`:

```
REPOSITORY                TAG                SIZE
foxglove_bridge_rolling   base (after)       1.03GB
foxglove_bridge_rolling   core (before)      945MB
ros                       rolling-ros-base   752MB
ros                       rolling-ros-core   474MB
ubuntu                    jammy              77.8MB
```
I did this as most of the large dependencies being manually installed should already be provided by the base tag. While the resulting image is a little larger `(1.03GB-945MB)=85MB`, the resulting size of the added layers to be regularly pushed over the network and saved to disk with every update to the bridge image is reduced from `(945MB-474MB)=471MB` to just `(1.03GB-752MB)=278MB`. Given 1) most robots using docker are more likely to have at least the image layers for the `ros-base` tag cached locally already for other running ROS containers, 2) the lower cadence of changed layers in the `ros-base` image, this should help reduce the bandwidth and storage requirements for deploying these bridge images.

### DIFF
--- a/Dockerfile.ros1
+++ b/Dockerfile.ros1
@@ -27,10 +27,11 @@ COPY ros1_foxglove_bridge src/ros-foxglove-bridge/ros1_foxglove_bridge
 # Build the Catkin workspace and ensure it's sourced when a container is run
 RUN . /opt/ros/$ROS_DISTRO/setup.sh \
   && catkin_make
-RUN echo "source devel/setup.bash" >> ~/.bashrc
 
-# Set the working folder at startup
-WORKDIR /ros1_ws
+# source workspace from entrypoint
+RUN sed --in-place \
+      's|^source .*|source "$ROS_WS/install/setup.bash"|' \
+      /ros_entrypoint.sh
 
 # Run foxglove_bridge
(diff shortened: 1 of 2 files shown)


achim-k (approved):
Looks good, thank you!

Dockerfile.ros1
```diff
@@ -1,32 +1,35 @@
...
+ENV ROS_WS /ros1_ws
+WORKDIR $ROS_WS
 
 # Add package.xml so we can install package dependencies.
-COPY package.xml /ros1_ws/src/ros-foxglove-bridge/package.xml
+COPY package.xml src/ros-foxglove-bridge/
 
 # Install rosdep dependencies
-RUN . /opt/ros/$ROS_DISTRIBUTION/setup.sh && \
+RUN . /opt/ros/$ROS_DISTRO/setup.sh && \
```
jhurliman, line 12: I don't see where `$ROS_DISTRO` is being set

PR #112, opened 2022-12-12
Add parameter support

### LINKED ISSUE #9: Parameters: get, set, list, subscribe, unsubscribe
For subscribing/unsubscribing to parameters, it may be helpful to support "subscribe all" / "unsubscribe all" since this is what Studio needs to drive the Parameters panel.

achim-k:
**Public-Facing Changes**
- Adds parameter support

**Description**
Implements support for parameters which was added to the ws-protocol specification in https://github.com/foxglove/ws-protocol/pull/288

This PR is rather big due to the different implementation of parameters for ROS 1 and ROS 2

ROS 1:
- Parameters are owned by ROS master
- Every node can get and set parameters on the master
- Simple API for setting / getting parameters
  - `ros::NodeHandle::getParamNames`, `ros::NodeHandle::getParam`, `ros::NodeHandle::setParam`

ROS 2 (see also [About-ROS-2-Parameters](https://docs.ros.org/en/rolling/Concepts/About-ROS-2-Parameters.html)):
- No single parameter store, parameters are owned by individual nodes 
- Getting / setting parameters of different nodes requires calling services for each involved node: 
  - `list_parameters`, `get_parameters`, `set_parameters`

Fixes #9

[FG-1203](https://linear.app/foxglove/issue/FG-1203/add-parameter-support)

### DIFF
--- a/ros2_foxglove_bridge/src/ros2_foxglove_bridge_node.cpp
+++ b/ros2_foxglove_bridge/src/ros2_foxglove_bridge_node.cpp
@@ -3,27 +3,32 @@
 int main(int argc, char* argv[]) {
   rclcpp::init(argc, argv);
 
-  auto dummyNode = std::make_shared<rclcpp::Node>("dummy");
-  auto numThreadsDescription = rcl_interfaces::msg::ParameterDescriptor{};
-  numThreadsDescription.name = "num_threads";
-  numThreadsDescription.type = rcl_interfaces::msg::ParameterType::PARAMETER_INTEGER;
-  numThreadsDescription.description =
-    "The number of threads to use for the ROS node executor. 0 means one thread per CPU core.";
-  numThreadsDescription.read_only = true;
-  numThreadsDescription.additional_constraints = "Must be a non-negative integer";
-  numThreadsDescription.integer_range.resize(1);
-  numThreadsDescription.integer_range[0].from_value = 0;
-  numThreadsDescription.integer_range[0].to_value = INT32_MAX;
-  numThreadsDescription.integer_range[0].step = 1;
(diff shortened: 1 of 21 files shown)


ros2_foxglove_bridge/src/ros2_foxglove_bridge_node.cpp
```diff
@@ -3,27 +3,31 @@
...
+    dummyNode->declare_parameter(numThreadsDescription.name, DEFAULT_NUM_THREADS,
+                                 numThreadsDescription);
+    numThreads = static_cast<size_t>(dummyNode->get_parameter(numThreadsDescription.name).as_int());
+  }
 
   auto executor =
     rclcpp::executors::MultiThreadedExecutor::make_shared(rclcpp::ExecutorOptions{}, numThreads);
 
-  rclcpp_components::ComponentManager componentManager(executor, "ComponentManager");
+  rclcpp_components::ComponentManager componentManager(executor, "foxglove_component_manager");
```
jhurliman, line 30: This may need to be `foxglove_bridge_component_manager` if it's supposed to be globally unique in the rosgraph, since we may introduce more Foxglove ROS nodes over time.

jhurliman (approved):

ros2_foxglove_bridge/src/ros2_foxglove_bridge_node.cpp
achim-k: Changed during rebasing. Since #125 we could actually get rid of the executable, since one can load the compoment directly in a launch file: https://docs.ros.org/en/humble/How-To-Guides/Launching-composable-nodes.html

PR #116, opened 2022-12-15
Skip sending messages if send buffer is full

achim-k:
**Public-Facing Changes**
- Prevent a connection's send buffer from getting to full when a client can't keep up with receiving messages

**Description**
- Allows the user to configure which topics shall be treated as best-effort topics for which messages can be dropped if the connection's send buffer reaches a certain limit. The added parameters are
  - **best_effort_max_buff_size**: Buffer size after which best-effort topic messages will be dropped
  - **best_effort_topics**: List of regex pattern for topic names which can be treated as best-effort (besides topics that are published with that qos)

### DIFF
--- a/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
@@ -31,6 +31,7 @@ using OpCode = websocketpp::frame::opcode::value;
 static const websocketpp::log::level APP = websocketpp::log::alevel::app;
 static const websocketpp::log::level RECOVERABLE = websocketpp::log::elevel::rerror;
 
+constexpr size_t MAX_SEND_BUFF_SIZE_BYTES = 100000000;  // 100MB
 constexpr uint32_t Integer(const std::string_view str) {
   uint32_t result = 0x811C9DC5;  // FNV-1a 32-bit algorithm
   for (char c : str) {
@@ -141,7 +142,7 @@ public:
   virtual void setClientMessageHandler(ClientMessageHandler handler) = 0;
 
   virtual void sendMessage(ConnHandle clientHandle, ChannelId chanId, uint64_t timestamp,
-                           std::string_view data) = 0;
+                           std::string_view data, bool bestEffort = false) = 0;
   virtual void broadcastTime(uint64_t timestamp) = 0;
 
   virtual std::optional<Tcp::endpoint> localEndpoint() = 0;
@@ -166,8 +167,9 @@ public:
   static bool USES_TLS;
 
   explicit Server(std::string name, LogCallback logger,
-                  const std::vector<std::string>& capabilities, const std::string& certfile = "",
-                  const std::string& keyfile = "");
+                  const std::vector<std::string>& capabilities,
+                  const size_t best_effort_buff_size_limit_kb = 1000UL,
+                  const std::string& certfile = "", const std::string& keyfile = "");
   virtual ~Server();
 
   Server(const Server&) = delete;
@@ -189,7 +191,7 @@ public:
   void setClientMessageHandler(ClientMessageHandler handler) override;
 
   void sendMessage(ConnHandle clientHandle, ChannelId chanId, uint64_t timestamp,
-                   std::string_view data) override;
+                   std::string_view data, bool bestEffort = false) override;
   void broadcastTime(uint64_t timestamp) override;
 
   std::optional<Tcp::endpoint> localEndpoint() override;
@@ -212,6 +214,7 @@ private:
   std::string _name;
   LogCallback _logger;
   std::vector<std::string> _capabilities;
+  size_t _best_effort_buff_size_limit_kb;
   std::string _certfile;
   std::string _keyfile;
   ServerType _server;
@@ -247,10 +250,12 @@ private:
 template <typename ServerConfiguration>
 inline Server<ServerConfiguration>::Server(std::string name, LogCallback logger,
                                            const std::vector<std::string>& capabilities,
+                                           const size_t best_effort_buff_size_limit_kb,
                                            const std::string& certfile, const std::string& keyfile)
     : _name(std::move(name))
     , _logger(logger)
     , _capabilities(capabilities)
+    , _best_effort_buff_size_limit_kb(best_effort_buff_size_limit_kb)
     , _certfile(certfile)
     , _keyfile(keyfile) {
   // Redirect logging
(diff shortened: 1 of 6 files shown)


ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -444,9 +444,11 @@ class FoxgloveBridge : public rclcpp::Node {
     }
 
     try {
+      const bool isReliableSubscription = qos.reliability() == rclcpp::ReliabilityPolicy::Reliable;
```
achim-k, line 447: This does not work well when testing this with a realsense bag file where all topics (even point clouds & raw image topics) are published with `ReliabilityPolicy::Reliable`

I believe we need indeed a parameter with which the user can control which topics must always be send

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -31,6 +31,7 @@ using OpCode = websocketpp::frame::opcode::value;
 static const websocketpp::log::level APP = websocketpp::log::alevel::app;
 static const websocketpp::log::level RECOVERABLE = websocketpp::log::elevel::rerror;
 
+constexpr size_t MAX_SEND_BUFF_SIZE_BYTES = 100000000;  // 100MB
```
jhurliman, line 34: This can be queried from websocketpp, we don't need to hardcode it

achim-k: Ah nice, will do

ros2_foxglove_bridge/include/foxglove_bridge/param_utils.hpp
```diff
@@ -0,0 +1,31 @@
...
+namespace foxglove_bridge {
+
+constexpr char PARAM_PORT[] = "port";
+constexpr char PARAM_ADDRESS[] = "address";
+constexpr char PARAM_USETLS[] = "tls";
+constexpr char PARAM_CERTFILE[] = "certfile";
+constexpr char PARAM_KEYFILE[] = "keyfile";
+constexpr char PARAM_MAX_QOS_DEPTH[] = "max_qos_depth";
+constexpr char PARAM_TOPIC_WHITELIST[] = "topic_whitelist";
+constexpr char PARAM_BEST_EFFORT_TOPICS[] = "best_effort_topics";
```
jhurliman, line 18: Although we should support both datatypes and topics at some point, datatypes will be a more common way to think about this I'm guessing since you generally want all of your sensor messages to be best-effort (for visualization at least), so all of the camera images and point clouds, but without having to name for example all six surround cameras and three lidars.

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
achim-k: I can retrieve the max message size (https://github.com/zaphoyd/websocketpp/blob/1b11fd301531e6df35a6107c1e8665b1e77a2d8e/docs/config.dox#L87) but so far I haven't found a way to get the max send buffer size. It's probably deeper down in asio

ros2_foxglove_bridge/include/foxglove_bridge/param_utils.hpp
achim-k: Should I add that? 

Usually these heavy topics have a common suffix, e.g. camera and point cloud topics. For instance, for the realsense bagfile I used the following patterns: `--ros-args -p best_effort_topics:=[".*points",".*image_raw"]`

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
jhurliman: Separately from deciding whether we need a parameter for topic reliability overrides, it sounds like there is a bug somewhere in the pipeline of recording MCAP data and playing it back in ROS 2 with the same QoS policy it was originally published with. Can you file a bug ticket with your repro steps?

ros2_foxglove_bridge/include/foxglove_bridge/param_utils.hpp
jhurliman: I think we should get to a place where this works out of the box for 99% of users without any custom configuration. In every example of poor streaming performance so far, the heavy topics have one of three datatypes: sensor_msgs/Image, sensor_msgs/CompressedImage, sensor_msgs/PointCloud2. If we're making the assertion that recording the correct QoS profile and using it for playback is broken and the correct fix for that is a workaround in foxglove_bridge, let's make that fix work out of the box by forcing best-effort on those three datatypes by default.

If we want to provide topic regex matching for power users, that's fine, but I don't see it as a high-priority requirement.

achim-k:
I think we should tackle this problem in steps as getting this right from the start seems difficult. I propose to do the following:

1. Do not send any messages anymore over a connection when the connection's send buffer has reached a certain threshold (user configurable). This ensures that the send buffer grows indefinitely until the kernel decides to kill the server.
2. Allow to configure the server such that reliable topics are treated differently. We must however still make sure that the send buffer does not grow indefinitely, so there should also be a limit for reliable topics
3. In the end only the client knows what channel is important to them, so we probably need to communicate that to the server (spec update)

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
achim-k: I did not record the mcap file myself, so I can't provide reproduction steps. The Mcap recording seems to be fine though. I believe that the realsense node advertised topics as reliable (not sure if that is the default)

For reference, below the output of `mcap list channels`: 

```
/rosout                                      reliability: 1  durability: 1
/tf_static                                   reliability: 1  durability: 1
/camera/aligned_depth_to_color/camera_info   reliability: 1  durability: 2
/camera/aligned_depth_to_color/color/points  reliability: 1  durability: 2
/camera/aligned_depth_to_color/image_raw     reliability: 1  durability: 2
/camera/depth/image_rect_raw                 reliability: 1  durability: 2
/camera/color/camera_info                    reliability: 1  durability: 2
/camera/depth/camera_info                    reliability: 1  durability: 2
/camera/color/image_raw                      reliability: 1  durability: 2
```

`reliability: 1` maps to [rclcpp::ReliabilityPolicy::Reliable](https://docs.ros2.org/galactic/api/rclcpp/namespacerclcpp.html#a4f037ff6e64986ad3e4ee883e0b2ed2d) (by default, the value of the first enumerator is zero if it is implicitly defined)

achim-k:
Closing in favor of #126

PR #124, opened 2022-12-19
Switch devcontainer to ROS2 humble

achim-k:
**Public-Facing Changes**
None

**Description**
ROS2 `galactic` is EOL since December 9th 2022. This PR changes the devcontainer to be based on ROS 2 `humble`. 

With that change, the devcontainer does not support ROS1 anymore. `humble` is supported on Ubuntu focal (Tier 3), but it needs to be compiled from source and I don't think it is worth the effort. Instead, for ROS1 development it is recommended to mount the workspace into a container and attach to the container with VsCode

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,36 +1,15 @@
-FROM ros:noetic-ros-core
+FROM ros:humble-ros-base
 
 # Install system dependencies
 RUN apt-get update && apt-get install -y --no-install-recommends \
-  build-essential \
   clang-format \
   curl \
-  git \
   git-lfs \
-  gnupg \
   lldb \
-  lsb-release \
   openssh-client \
-  python3-colcon-common-extensions \
-  python3-pip \
-  python3-rosdep \
-  python3-rosinstall \
-  python3-rosinstall-generator \
-  python3-wstool \
+  ros-$ROS_DISTRO-foxglove-msgs \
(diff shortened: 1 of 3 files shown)


jtbandes (approved):

jhurliman (changes requested):
> Instead, for ROS1 development it is recommended to mount the workspace into a container and attach to the container with VsCode

I don't see any instructions in the README on what this means or how to do it

jhurliman:
I'm -1 on this change. The current setup of a single container for ROS 1 and ROS 2 makes it very convenient to develop for both platforms. Migrating off the EOL Galactic is a nice goal, but if the cost is a hampered workflow, I don't think it's worth it.

achim-k:
> I'm -1 on this change. The current setup of a single container for ROS 1 and ROS 2 makes it very convenient to develop for both platforms. Migrating off the EOL Galactic is a nice goal, but if the cost is a hampered workflow then I don't think it's worth it.

What do you propose? Continue with the existing devcontainer?

jhurliman:
> > I'm -1 on this change. The current setup of a single container for ROS 1 and ROS 2 makes it very convenient to develop for both platforms. Migrating off the EOL Galactic is a nice goal, but if the cost is a hampered workflow then I don't think it's worth it.
> 
> What do you propose? Continue with the existing devcontainer?

Yes

PR #126, opened 2022-12-23
Drop messages when connection send buffer limit has been reached

achim-k:
**Public-Facing Changes**
- Drop messages when connection send buffer limit has been reached

**Description**
We currently do have the problem that slow clients/connections can cause the server's connection send buffer to grow indefinitely potentially leading to the server being killed as it consumes too much memory. This PR introduces a user configurable send buffer limit. If the limit has been reached, messages that are to be send over this connection will be dropped in order to prevent the send buffer growing indefinitely.

See also #116

### DIFF
--- a/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
@@ -31,6 +31,8 @@ using OpCode = websocketpp::frame::opcode::value;
 static const websocketpp::log::level APP = websocketpp::log::alevel::app;
 static const websocketpp::log::level RECOVERABLE = websocketpp::log::elevel::rerror;
 
+constexpr size_t DEFAULT_SEND_BUFFER_LIMIT_BYTES = 10000000UL;  // 10 MB
+
 constexpr uint32_t Integer(const std::string_view str) {
   uint32_t result = 0x811C9DC5;  // FNV-1a 32-bit algorithm
   for (char c : str) {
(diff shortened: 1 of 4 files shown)


foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -800,6 +806,20 @@ inline void Server<ServerConfiguration>::broadcastChannels() {
...
+  const auto con = _server.get_con_from_hdl(clientHandle, ec);
+  if (ec || !con) {
+    return;
+  }
+
+  const auto bufferSizeinBytes = con->get_buffered_amount();
+  if (bufferSizeinBytes >= _send_buffer_limit_bytes) {
+    _server.get_elog().write(RECOVERABLE,
+                             "Send buffer for client '" + remoteEndpointString(clientHandle) +
+                               "' is full, dropping message on channel " + std::to_string(chanId));
```
achim-k, line 819: We may need log throttling here...

jhurliman (approved):

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -800,6 +807,20 @@ inline void Server<ServerConfiguration>::broadcastChannels() {
...
+  const auto con = _server.get_con_from_hdl(clientHandle, ec);
+  if (ec || !con) {
+    return;
+  }
+
+  const auto bufferSizeinBytes = con->get_buffered_amount();
+  if (bufferSizeinBytes >= _send_buffer_limit_bytes) {
+    _server.get_elog().write(WARNING,
+                             "Send buffer for client '" + remoteEndpointString(clientHandle) +
+                               "' is full, dropping message on channel " + std::to_string(chanId));
```
jhurliman, line 820: Remove this

PR #128, opened 2023-01-04
Add throttled logging when send buffer is full

achim-k:
**Public-Facing Changes**
- Add logging when a connection's send buffer is full

**Description**
Logs a warning when a connection's send buffer limit has been reached. To avoid log spam, we log this message at max. every 2.5 seconds.

### DIFF
--- a/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
@@ -23,18 +23,16 @@
 #include "websocket_tls.hpp"
 
 // Debounce a function call (tied to the line number)
-// This macro takes in a function, the debounce time in milliseconds, and any
-// arguments to pass to the function.
-#define FOXGLOVE_DEBOUNCE(f, ms, ...)                                                     \
-  do {                                                                                    \
-    static auto last_call_##__LINE__ = std::chrono::system_clock::now();                  \
-    const auto now = std::chrono::system_clock::now();                                    \
-    if (std::chrono::duration_cast<std::chrono::milliseconds>(now - last_call_##__LINE__) \
-          .count() > ms) {                                                                \
-      last_call_##__LINE__ = now;                                                         \
-      f(__VA_ARGS__);                                                                     \
-    }                                                                                     \
-  } while (false)
+// This macro takes in a function and the debounce time in milliseconds
+#define FOXGLOVE_DEBOUNCE(f, ms)                                                               \
+  {                                                                                            \
+    static auto last_call = std::chrono::system_clock::now();                                  \
+    const auto now = std::chrono::system_clock::now();                                         \
+    if (std::chrono::duration_cast<std::chrono::milliseconds>(now - last_call).count() > ms) { \
+      last_call = now;                                                                         \
+      f();                                                                                     \
+    }                                                                                          \
+  }
 
 namespace foxglove {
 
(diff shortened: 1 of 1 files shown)


foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -21,6 +22,20 @@
 #include "websocket_notls.hpp"
 #include "websocket_tls.hpp"
 
+// Debounce a function call (tied to the line number)
+// This macro takes in a function, the debounce time in milliseconds, and any
+// arguments to pass to the function.
+#define FOXGLOVE_DEBOUNCE(f, ms, ...)                                                     \
+  do {                                                                                    \
+    static auto last_call_##__LINE__ = std::chrono::system_clock::now();                  \
```
jtbandes, line 30: I don't think `__LINE__` is really necessary since the variable is scoped within the do{}while() anyway.

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -815,6 +830,11 @@ inline void Server<ServerConfiguration>::sendMessage(ConnHandle clientHandle, Ch
 
   const auto bufferSizeinBytes = con->get_buffered_amount();
   if (bufferSizeinBytes >= _send_buffer_limit_bytes) {
+    FOXGLOVE_DEBOUNCE(
+      [this](websocketpp::log::level level, const std::string& msg) {
+        _server.get_elog().write(level, msg);
+      },
```
jtbandes, line 836: why not just `[&] { _server.get_elog().write(WARNING, "..."); }` ?

jtbandes (approved):

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
achim-k: Without `__LINE__`, if you use the macro in different places, they would interfere since they would use the same static variable

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
achim-k: That's what I had initially, this gave however errors for melodic:
```
ISO C++11 requires at least one argument for the ...'' in a variadic macro
```

jtbandes: I suppose you could just not allow variadic args.

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
jtbandes: https://godbolt.org/z/MWasY7on7

jhurliman: @achim-k that would be true if this was a function, but it's a macro. Each instance of `FOXGLOVE_DEBOUNCE` will copy-paste a new variable definition inside a `do { }` block.

achim-k: Oh, you are right! Will change it

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
achim-k: Will remove it

jtbandes (approved):

PR #131, opened 2023-01-10
Communicate supported encodings for client-side publishing

achim-k:
**Public-Facing Changes**
- Communicate supported encodings for client-side publishing

**Description**
Corresponding spec PR: https://github.com/foxglove/ws-protocol/pull/326

> A client currently has no way in determining which message encodings the server supports for client publishing. This PR adds a field to the serverInfo message which informs clients about supported encodings.

### DIFF
--- a/ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
+++ b/ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
@@ -62,14 +62,16 @@ public:
     if (_useSimTime) {
       serverCapabilities.push_back(foxglove::CAPABILITY_TIME);
     }
+    const std::vector<std::string> supportedEncodings = {"cdr"};
 
     if (useTLS) {
       _server = std::make_unique<foxglove::Server<foxglove::WebSocketTls>>(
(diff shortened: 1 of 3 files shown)


ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -62,14 +62,16 @@ class FoxgloveBridge : public rclcpp::Node {
     if (_useSimTime) {
       serverCapabilities.push_back(foxglove::CAPABILITY_TIME);
     }
+    const std::vector<std::string> supportedEncodings = {"cdr"};
```
jtbandes, line 65: why doesn't this one use a static constant when the other one does?

jtbandes (approved):

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
achim-k: I guess because we originally planned to support both `cdr` and `idl`. Will change it in the future when we have some time for cleanups

PR #134, opened 2023-01-12
Send ROS_DISTRO to clients via metadata field

achim-k:
**Public-Facing Changes**
- Send current `$ROS_DISTRO` to clients via `serverInfo`'s `metadata` field

**Description**
- Add support for `serverInfo`'s `metadata` field (https://github.com/foxglove/ws-protocol/pull/334)
- Inform clients about the value of the `$ROS_DISTRO` environment variable via the `metadata` field

Blocked by https://github.com/foxglove/ws-protocol/pull/334

### DIFF
--- a/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
@@ -194,6 +194,7 @@ public:
   explicit Server(std::string name, LogCallback logger,
                   const std::vector<std::string>& capabilities,
                   const std::vector<std::string>& supportedEncodings = {},
+                  const std::unordered_map<std::string, std::string>& metadata = {},
                   size_t send_buffer_limit_bytes = DEFAULT_SEND_BUFFER_LIMIT_BYTES,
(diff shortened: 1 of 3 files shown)


foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -194,6 +194,7 @@ class Server final : public ServerInterface {
   explicit Server(std::string name, LogCallback logger,
                   const std::vector<std::string>& capabilities,
                   const std::vector<std::string>& supportedEncodings = {},
+                  const std::unordered_map<std::string, std::string>& metadata = {},
```
achim-k, line 197: For simplicity, I decided to make this a map of string-string pairs. When e.g. using `std::variant`, we would have to provide json conversions for it.

jhurliman: I made an indifferent comment on the spec PR, but after seeing this I think we should change the spec to `<string, string>`

jhurliman (approved):

PR #141, opened 2023-01-24
Fix flaky parameter test (noetic)

achim-k:
**Public-Facing Changes**
- Fixes flaky parameter test (noetic)

**Description**
Instead of using the cached value, we now always query the ROS master for the latest parameter value.

I am actually not sure why it does not work when using [getParamCached](http://docs.ros.org/en/noetic/api/roscpp/html/classros_1_1NodeHandle.html#a29e8424c7760136ad5b9294ed1de84e7). I propose to merge this PR now and investigate using `getParamCached` later on again


jhurliman:
Merge main in and the remaining CI flakes should go away

jhurliman (approved):

PR #144, opened 2023-02-01
Link to docs on the Foxglove website

esthersweon:
Per @amacneil 's request.

**Public-Facing Changes**

- Remove docs from README
- Link to docs on the Foxglove website
- Blocked by https://github.com/foxglove/website/pull/971

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,90 +15,9 @@ Live debugging of ROS systems has traditionally relied on running ROS tooling su
 
 The `foxglove_bridge` uses the [Foxglove WebSocket protocol](https://github.com/foxglove/ws-protocol), a similar protocol to rosbridge but with the ability to support additional schema formats such as ROS 2 `.msg` and ROS 2 `.idl`, parameters, graph introspection, and non-ROS systems. The bridge is written in C++ and designed for high performance with low overhead to minimize the impact to your robot stack.
 
-## Installation
+## Documentation
 
-The `foxglove_bridge` package is available for ROS 1 Melodic and Noetic, and ROS 2 Humble and Rolling. Earlier releases of ROS will not be supported due to API design and/or performance limitations. The package can be installed with the following command:
-
-```bash
-$ sudo apt install ros-$ROS_DISTRO-foxglove-bridge
-```
-
-## Running the bridge
-
-To run the bridge node, it is recommended to use the provided launch file:
-
-**ROS 1**
-
-```bash
-$ roslaunch foxglove_bridge foxglove_bridge.launch --screen
-```
-
-**ROS 2**
-
-```bash
-$ ros2 launch foxglove_bridge foxglove_bridge_launch.xml
-```
-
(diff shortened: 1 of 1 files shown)


defunctzombie:
@amacneil What's the reasoning here? As a developer looking to use a package like this I often find the README a pretty good place for docs. Granted some projects do decide to host docs separately from the README.

achim-k:
> @amacneil What's the reasoning here? As a developer looking to use a package like this I often find the README a pretty good place for docs. Granted some projects do decide to host docs separately from the README.

I agree, I also find the README helpful and I would not remove it. The big advantage is versioning:
Documentation updates is linked to code changes & releases. For instance, https://index.ros.org/p/foxglove_bridge does not show the latest README but only the version of the latest releases. I believe this is also what we should do for the foxglove-bridge docs that are to be hosted on the foxglove website. I see little value in also documenting unreleased foxglove-bridge features on the foxglove website.

README.md
```diff
@@ -15,89 +15,35 @@ Live debugging of ROS systems has traditionally relied on running ROS tooling su
...
-## Clients
-
-[Foxglove Studio](https://foxglove.dev/studio) connects to `foxglove_bridge` for live robotics visualization.
-
-### Building from source
-
-You can also try `foxglove_bridge` by building from source or running a pre-built Docker container.
-
-#### Fetch source and install dependencies
+To build `foxglove_bridge` from source, you must first fetch your source and install dependencies:
```
defunctzombie, line 24:
```suggestion
To build `foxglove_bridge` from source, you should clone the source to your ROS workspace and install dependencies:
```

defunctzombie (approved):

jhurliman (changes requested):
~Net negative change~

EDIT: Achim wrote up a pretty good description of why erasing the README for a ROS package is not a good idea. I'm marking this as "changes requested" to hold up the PR for discussion.

amacneil:
Two anti patterns I would like us to avoid:
- Having multiple sets of documentation for the same thing. It's fine for the readme to be a "jumping off point", but it should send you to the canonical documentation.
- Forcing people to jump between multiple docs sites and readme's to solve a problem. For example, before we started cleaning up our readme's, it was necessary to jump between `ws-protocol`, `foxglove/schemas`, and now `ros-foxglove-bridge` to gain all the context on how to connect your robot to studio. I should go to one place for this information and not need to find my way through a web of github repos.

So it's fine to have some basic info in the readme but let's point people to the canonical docs.

defunctzombie:
> Net negative change

@jhurliman can you please be more specific? This type of comment is not helpful since there is no clear action to take in response to your change request.

jhurliman:
Discussed in Slack, we are going to rework the docs approach to keep per-version specific instructions such as launchfile instructions and parameter descriptions in git while maintaining high-level overview and links to instructions on the foxglove.dev/docs site.

PR #145, opened 2023-02-03
Allow unsetting (deleting) parameters

achim-k:
**Public-Facing Changes**
- Allow unsetting (deleting) parameters

**Description**
Implements the specification update in https://github.com/foxglove/ws-protocol/pull/354

### DIFF
--- a/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
@@ -974,7 +974,14 @@ template <typename ServerConfiguration>
 inline void Server<ServerConfiguration>::publishParameterValues(
   ConnHandle hdl, const std::vector<Parameter>& parameters,
   const std::optional<std::string>& requestId) {
-  nlohmann::json jsonPayload{{"op", "parameterValues"}, {"parameters", parameters}};
+  // Filter out parameters which are not set.
+  std::vector<Parameter> nonEmptyParameters;
+  std::copy_if(parameters.begin(), parameters.end(), std::back_inserter(nonEmptyParameters),
+               [](const auto& p) {
+                 return p.getType() != ParameterType::PARAMETER_NOT_SET;
+               });
+
+  nlohmann::json jsonPayload{{"op", "parameterValues"}, {"parameters", nonEmptyParameters}};
   if (requestId) {
     jsonPayload["id"] = requestId.value();
   }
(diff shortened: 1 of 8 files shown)


foxglove_bridge_base/src/serialization.cpp
```diff
@@ -21,14 +21,20 @@ void to_json(nlohmann::json& j, const Parameter& p) {
...
   j["name"] = p.getName();
 }
 
 void from_json(const nlohmann::json& j, Parameter& p) {
   const auto name = j["name"].get<std::string>();
+
+  if (j.find("value") == j.end()) {
+    p = Parameter(name);  // Value is not set (undefined).
+    return;
+  }
```
achim-k, line 36: I think it would be more explicit if the value was set to `null` (nlohmann::detail::value_t::boolean)

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -974,7 +974,14 @@ template <typename ServerConfiguration>
 inline void Server<ServerConfiguration>::publishParameterValues(
   ConnHandle hdl, const std::vector<Parameter>& parameters,
   const std::optional<std::string>& requestId) {
-  nlohmann::json jsonPayload{{"op", "parameterValues"}, {"parameters", parameters}};
+  // Filter out parameters which are not set.
```
achim-k, line 977: The spec currently does not declare what parameter message to expect when requesting a non-existing / deleted parameter. Should it be
```json
{
  "op": "parameterValues", 
  "parameters": []
}
```
or
```json
{
  "op": "parameterValues", 
  "parameters": [{ "name": "/deleted_param", "value": undefined }]
}
```

foxglove_bridge_base/src/serialization.cpp
jhurliman: I don't understand the comment here, `null` and undefined are not the same thing.

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
jhurliman: The second example is not valid JSON. There is no `undefined` literal in https://www.rfc-editor.org/rfc/rfc8259

achim-k: Adapted my first comment by removing the `value` field

jhurliman: I think the first is better?

achim-k: OK, that's also how it is currently implemented

achim-k:
@jhurliman OK for you to merge as is?

jhurliman (approved):

PR #152, opened 2023-02-08
Add option to use permessage-deflate compression

achim-k:
**Public-Facing Changes**
- Add option to use permessage-deflate compression

**Description**
- Add option to use permessage-deflate compression

Found out that this was possible when implementing #151. Not sure if it is really useful tho, the additional CPU load is quite high (I can provide some measurements on request)


rgov:
Compression would definitely be useful from my perspective. My systems are running on limited cellular or satellite connections and I'd rather spend CPU time to save bandwidth.

Might be cool if it could determine automatically that a certain topic doesn't compress very well (i.e., image data).

jhurliman (approved):

PR #165, opened 2023-02-16
Run client handler functions in separate thread

### LINKED ISSUE #164: [ROS2] Custom types and parameters fail
**Description**
Right now, we have a custom service type that we do not really care about. The problem is that the bridge tries to get it and since it does not recognise the type it throws errors on the terminal as well as warnings on the ros log. While this happens we lose messages for about 3 seconds until it comes back to life.

Log window:
image

- Version: Latest main
- Platform: Ubuntu 22.04 (ROS humble)

achim-k:
**Public-Facing Changes**
- Run client handlers in separate thread

**Description**
Builds on top of #156 

So far we have executed client handler functions in the websocket server thread. This could cause issues with clients not receiving messages anymore while a longer-running handler function was executing. This PR fixes this by handler functions being executed in a separate thread.

Fixes #164 
Fixes FG-2029

### DIFF
--- a/foxglove_bridge_base/include/foxglove_bridge/common.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/common.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <cstring>
 #include <stdint.h>
 #include <string>
 #include <vector>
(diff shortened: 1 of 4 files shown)


foxglove_bridge_base/include/foxglove_bridge/common.hpp
```diff
@@ -74,26 +75,28 @@ struct ClientMessage {
...
-                const ClientAdvertisement& advertisement, size_t dataLength, const uint8_t* data)
+                const ClientAdvertisement& advertisement, size_t dataLength, const uint8_t* rawData)
       : logTime(logTime)
       , publishTime(publishTime)
       , sequence(sequence)
       , advertisement(advertisement)
       , dataLength(dataLength)
-      , data(data) {}
+      , data(dataLength) {
+    std::memcpy(data.data(), rawData, dataLength);
```
jhurliman, line 90: Is the additional copy worth it? We should test on a low-power platform like a Jetson Nano.

foxglove_bridge_base/include/foxglove_bridge/common.hpp
```diff
@@ -74,26 +75,28 @@ struct ClientMessage {
...
-  const uint8_t* data;
+  std::vector<uint8_t> data;
 
   ClientMessage(uint64_t logTime, uint64_t publishTime, uint32_t sequence,
-                const ClientAdvertisement& advertisement, size_t dataLength, const uint8_t* data)
+                const ClientAdvertisement& advertisement, size_t dataLength, const uint8_t* rawData)
       : logTime(logTime)
       , publishTime(publishTime)
       , sequence(sequence)
       , advertisement(advertisement)
```
jhurliman, line 87: This is now a copy, but I'm assuming `ClientAdvertisement` is relative cheap to copy?

foxglove_bridge_base/include/foxglove_bridge/common.hpp
achim-k: We have to do a copy now since we call the handler in a different thread and the pointer to the original data is likely not valid anymore.

foxglove_bridge_base/include/foxglove_bridge/common.hpp
achim-k: Yes I think so too. At this stage I would rather have an extra copy then some hard to find bug. But yes, there is definitely potential for improvement later on

jhurliman (approved):

PR #167, opened 2023-02-17
Allow subscribing to connection graph updates

achim-k:
**Public-Facing Changes**
- Allow subscribing to connection graph updates (https://github.com/foxglove/ws-protocol/pull/368)

**Description**
Implements https://github.com/foxglove/ws-protocol/pull/368

This PR will allow using the TopicGraph panel in Foxglove Studio

### DIFF
--- a/foxglove_bridge_base/include/foxglove_bridge/server_interface.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/server_interface.hpp
@@ -14,7 +14,7 @@ namespace foxglove {
 
 constexpr size_t DEFAULT_SEND_BUFFER_LIMIT_BYTES = 10000000UL;  // 10 MB
 
-typedef std::unordered_map<std::string, std::unordered_set<std::string>> MapOfSets;
+using MapOfSets = std::unordered_map<std::string, std::unordered_set<std::string>>;
 
 struct ServerOptions {
   std::vector<std::string> capabilities;
--- a/foxglove_bridge_base/include/foxglove_bridge/server_interface.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/server_interface.hpp
@@ -4,6 +4,7 @@
 #include <optional>
 #include <string>
 #include <unordered_map>
+#include <unordered_set>
 #include <vector>
 
 #include "common.hpp"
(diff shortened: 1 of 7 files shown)


jhurliman (approved):

foxglove_bridge_base/include/foxglove_bridge/server_interface.hpp
```diff
@@ -13,6 +14,8 @@ namespace foxglove {
 
 constexpr size_t DEFAULT_SEND_BUFFER_LIMIT_BYTES = 10000000UL;  // 10 MB
 
+typedef std::unordered_map<std::string, std::unordered_set<std::string>> MapOfSets;
```
jtbandes, line 17: nit: I personally prefer `using` everywhere since it feels more readable
```suggestion
using MapOfSets = std::unordered_map<std::string, std::unordered_set<std::string>>;
```

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -446,6 +492,14 @@ class FoxgloveBridge : public rclcpp::Node {
       std::bind(&FoxgloveBridge::serviceRequest, this, request, hdl));
   }
 
+  void subscribeConnectionGraphHandler(bool subscribe) {
+    if (_subscribeGraphUpdates = subscribe) {
```
jtbandes, line 496: Shouldn't this be a warning/error? Maybe we don't have the standard warning flags enabled? 😱 
```suggestion
    if ((_subscribeGraphUpdates = subscribe)) {
```

jtbandes (approved):

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
achim-k: What warning would that be?

> Maybe we don't have the standard warning flags enabled?

I'm pretty sure we do

jtbandes: Clang seems to warn on this with default flags: https://godbolt.org/z/n5Pz67aaP

It's called `-Wparentheses`

achim-k: There seems to be something wrong with how compiler flags are defined

https://github.com/foxglove/ros-foxglove-bridge/blob/f4b2f68cde3cc87ad65cb95a36cd7690da34b3a5/CMakeLists.txt#L41-L49

I will investigate this when I have time. I read somewhere that tinkering with `CMAKE_CXX_FLAGS` is not the recommended way to do it.

achim-k: `-Wparentheses` is included in `-Wall` but gcc does not detect it in this specific case. Clang does, so I think I will also enable clang as the default compiler for CI

PR #172, opened 2023-02-22
Avoid re-advertising existing channels when advertising new channels

### LINKED ISSUE #146: Broadcast only newly added channels instead of all existing channels
**Description**
When new channels are added, we currently broadcast all channels (including the newly added ones) instead of just the added channels

https://github.com/foxglove/ros-foxglove-bridge/blob/d29159c10127ecfac9262ddaa657def371f1aeab/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp#L952-L971

One way to achieve this (without having the server remember which channels were newly added) is to change the API to

```cpp
std::vector<ChannelId> addChannels(const std::vector<ChannelWithoutId>& channels);
void removeServices(const std::vector<ChannelId>& channelIds);
```

achim-k:
**Public-Facing Changes**
- Avoid re-advertising existing channels when advertising new channels

**Description**
When new channels are advertised, we broadcast all available channels to clients rather than just the newly added ones. This PR fixes it.

Fixes #146
Fixes [FG-1793](https://linear.app/foxglove/issue/FG-1793/broadcast-only-newly-added-channels-instead-of-all-existing-channels)

### DIFF
--- a/foxglove_bridge_base/include/foxglove_bridge/server_interface.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/server_interface.hpp
@@ -51,9 +51,8 @@ public:
   virtual void start(const std::string& host, uint16_t port) = 0;
   virtual void stop() = 0;
 
-  virtual ChannelId addChannel(ChannelWithoutId channel) = 0;
-  virtual void removeChannel(ChannelId chanId) = 0;
-  virtual void broadcastChannels() = 0;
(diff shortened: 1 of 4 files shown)


achim-k:
Pinging for review @jhurliman @jtbandes

jhurliman (approved):
Can you port this change to `ws-protocol` repo too?

achim-k:
> Can you port this change to `ws-protocol` repo too?

Already done. I have ported the version of this PR

PR #181, opened 2023-03-07
Add parameter `client_topic_whitelist` for whitelisting client-published topics

### LINKED ISSUE #180: [ROS1] Topic whitelist is not bidirectional
**Description**

We are evaluating using `foxglove_bridge` for forwarding data from a ROS1 stack that serves as a "gateway" for communicating with our robot to remote clients running Foxglove Studio. 

One feature that is especially important to us is the topic whitelist, since we want the ability to selectively control how much data gets sent from the gateway stack to the clients. However, we also want to limit the publishing capabilities from client to gateway. This is necessary because we have custom buttons in our Foxglove Studio layout that send commands to the robot (abort mission, change configuration, etc.) and we'd like to have a "read-only" mode for our UI that makes these buttons non-functional. We thought we could accomplish this via the topic whitelist by simply black-listing the topics that contain the control output from these button presses. However, in testing the bridge we found that the whitelist is not bi-directional, so publishing requests from client to the websocket server are un-guarded and make it through regardless of any specified topic blacklists. 

We made a patch for this issue that makes the whitelist bi-directional and are currently testing it (initial results indicate that it does work for blacklisting the publishes from our button presses). Is there a design reason that the whitelist is not currently configured to be bi-directional? If not, we'd be happy to open a PR to upstream the patch.

- Version: latest source
- Platform: ROS1 Melodic, Ubuntu 18.04, ARM device

achim-k:
### Public-Facing Changes
- Add parameter `client_topic_whitelist` for whitelisting client-published topics

### Description
Fixes #180

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ To run the bridge node, it is recommended to use the provided launch file:
 **ROS 1**
 
 ```bash
-$ roslaunch --screen foxglove_bridge foxglove_bridge.launch port:=8765
+$ roslaunch --screen foxglove_bridge foxglove_bridge.launch port:=8765 # topic_whitelist:=[/foo/.*,/bar/baz]
 ```
 
 ```xml
@@ -48,7 +48,7 @@ $ roslaunch --screen foxglove_bridge foxglove_bridge.launch port:=8765
 **ROS 2**
 
 ```bash
-$ ros2 launch foxglove_bridge foxglove_bridge_launch.xml port:=8765
+$ ros2 launch foxglove_bridge foxglove_bridge_launch.xml port:=8765 # topic_whitelist:=[/foo/.*,/bar/baz]
 ```
 
 ```xml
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -32,7 +32,7 @@ To run the bridge node, it is recommended to use the provided launch file:
 **ROS 1**
 
 ```bash
-$ roslaunch --screen foxglove_bridge foxglove_bridge.launch port:=8765
+$ roslaunch --screen foxglove_bridge foxglove_bridge.launch port:=8765 # topic_whitelist:=[/foo/.*,/bar/baz]
```
jtbandes, line 35: Comment syntax could be confusing

jtbandes (approved):
Seems ok, but consider making it a separate whitelist to avoid breaking changes

jtbandes (approved):

PR #186, opened 2023-03-08
Add support for `schemaEncoding` field

### LINKED ISSUE #8: Support IDL message definitions
Right now we only specify message encoding, not schema encoding so it's not possible to distinguish between ros2msg and ros2idl.

achim-k:
### Public-Facing Changes

- Add support for `schemaEncoding` field

### Description
Implements the specification update in https://github.com/foxglove/ws-protocol/pull/385
Fixes #8

### DIFF
--- a/foxglove_bridge_base/src/serialization.cpp
+++ b/foxglove_bridge_base/src/serialization.cpp
@@ -10,14 +10,20 @@ void to_json(nlohmann::json& j, const Channel& c) {
     {"schemaName", c.schemaName},
     {"schema", c.schema},
   };
+
+  if (c.schemaEncoding.has_value()) {
+    j["schemaEncoding"] = c.schemaEncoding.value();
+  }
 }
 void from_json(const nlohmann::json& j, Channel& c) {
-  ChannelWithoutId channelWithoutId{
-    j["topic"].get<std::string>(),
-    j["encoding"].get<std::string>(),
-    j["schemaName"].get<std::string>(),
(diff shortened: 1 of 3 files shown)


foxglove_bridge_base/src/serialization.cpp
```diff
@@ -10,14 +10,20 @@ void to_json(nlohmann::json& j, const Channel& c) {
...
-  ChannelWithoutId channelWithoutId{
-    j["topic"].get<std::string>(),
-    j["encoding"].get<std::string>(),
-    j["schemaName"].get<std::string>(),
-    j["schema"].get<std::string>(),
-  };
+  const auto schemaEncoding =
+    j.find("schemaEncoding") == j.end()
+      ? std::optional<std::string>(std::nullopt)
+      : std::optional<std::string>(j["schemaEncoding"].get<std::string>());
```
jtbandes, line 22: nit
```suggestion
  const std::optional<std::string> schemaEncoding =
    j.find("schemaEncoding") == j.end()
      ? std::nullopt
      : j["schemaEncoding"].get<std::string>();
```

jtbandes (approved):
🎉

foxglove_bridge_base/src/serialization.cpp
achim-k: Also initially tried that, but CPP is rather strict and complains that the types in the branches (not sure what the correct term is) differs.

PR #191, opened 2023-03-09
Add more exception handling

achim-k:
### Public-Facing Changes
- Add more exception handling

### Description
Adds more exception handling in places where it is unlikely but not impossible that exceptions are being thrown

### DIFF
--- a/ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
+++ b/ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
@@ -449,7 +449,13 @@ private:
         _server->remoteEndpointString(clientHandle).c_str(), channelId, clientPublications.size());
       return;
     }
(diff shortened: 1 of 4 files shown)


jtbandes (approved):

ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
```diff
@@ -449,7 +449,13 @@ class FoxgloveBridge : public nodelet::Nodelet {
...
       return;
     }
-    channelPublicationIt->second.publish(msg);
+
+    try {
+      channelPublicationIt->second.publish(msg);
+    } catch (const std::exception& ex) {
+      ROS_ERROR("Failed to publish message on topic '%s'",
+                channelPublicationIt->second.getTopic().c_str());
+    }
```
achim-k, line 458: The exception parameter `ex` is not used and should have caused CI to fail!

PR #199, opened 2023-03-15
Add support for byte array params

### LINKED ISSUE #198: ROS2: Parameters of type PARAMETER_BYTE_ARRAY are not supported
**Description**
Byte array parameters are currently not supported and the exception `Unsupported parameter type` is raised when such a parameter is encountered.

**Expected Behavior**
Byte array parameters should be supported

achim-k:
### Public-Facing Changes
- Add support for byte array params [ROS2]

### Description
Adds support for byte array parameters according to the ws-protocol specification update made in https://github.com/foxglove/ws-protocol/pull/396

Fixes #198

### DIFF
--- a/foxglove_bridge_base/src/parameter.cpp
+++ b/foxglove_bridge_base/src/parameter.cpp
@@ -38,6 +38,11 @@ Parameter::Parameter(const std::string& name, std::string value)
     , _type(ParameterType::PARAMETER_STRING)
     , _value(value) {}
 
+Parameter::Parameter(const std::string& name, std::string value, ParameterType type)
+    : _name(name)
+    , _type(type)
+    , _value(value) {}
+
 Parameter::Parameter(const std::string& name, const std::vector<bool>& value)
     : _name(name)
     , _type(ParameterType::PARAMETER_BOOL_ARRAY)
--- a/foxglove_bridge_base/src/parameter.cpp
+++ b/foxglove_bridge_base/src/parameter.cpp
@@ -38,9 +38,9 @@ Parameter::Parameter(const std::string& name, std::string value)
     , _type(ParameterType::PARAMETER_STRING)
     , _value(value) {}
 
-Parameter::Parameter(const std::string& name, std::string value, ParameterType type)
+Parameter::Parameter(const std::string& name, const std::vector<unsigned char>& value)
     : _name(name)
-    , _type(type)
+    , _type(ParameterType::PARAMETER_BYTE_ARRAY)
     , _value(value) {}
 
 Parameter::Parameter(const std::string& name, const std::vector<bool>& value)
--- a/foxglove_bridge_base/src/serialization.cpp
+++ b/foxglove_bridge_base/src/serialization.cpp
@@ -39,8 +39,9 @@ void to_json(nlohmann::json& j, const Parameter& p) {
   } else if (paramType == ParameterType::PARAMETER_STRING) {
     j["value"] = p.getValue<std::string>();
   } else if (paramType == ParameterType::PARAMETER_BYTE_ARRAY) {
-    const auto paramValue = p.getValue<std::vector<unsigned char>>();
-    const std::string strValue(paramValue.begin(), paramValue.end());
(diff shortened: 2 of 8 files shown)


foxglove_bridge_base/src/parameter.cpp
```diff
@@ -38,6 +38,11 @@ Parameter::Parameter(const std::string& name, std::string value)
     , _type(ParameterType::PARAMETER_STRING)
     , _value(value) {}
 
+Parameter::Parameter(const std::string& name, std::string value, ParameterType type)
+    : _name(name)
+    , _type(type)
+    , _value(value) {}
```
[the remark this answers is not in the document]
jtbandes: Should this constructor assert that the type is either string or byte array?

Alternatively, should we change the constructor to accept a `std::span<std::byte>` or similar and do the base64 conversion internally?

This page has some interesting advice, seems to suggest strong typing for this type of use case rather than tag dispatching (which I was considering recommending): https://www.fluentcpp.com/2018/04/27/tag-dispatching/

ros2_foxglove_bridge/src/parameter_interface.cpp
```diff
@@ -33,6 +80,8 @@ static rclcpp::Parameter toRosParam(const foxglove::Parameter& p) {
     return rclcpp::Parameter(p.getName(), p.getValue<double>());
   } else if (paramType == ParameterType::PARAMETER_STRING) {
     return rclcpp::Parameter(p.getName(), p.getValue<std::string>());
+  } else if (paramType == ParameterType::PARAMETER_BYTE_ARRAY) {
+    return rclcpp::Parameter(p.getName(), base64Decode(p.getValue<std::string>()));
```
jtbandes, line 84: What if base64 decoding fails (contains invalid chars)?

ros2_foxglove_bridge/src/parameter_interface.cpp
```diff
@@ -20,6 +20,53 @@ static std::string prependNodeNameToParamName(const std::string& paramName,
   return nodeName + PARAM_SEP + paramName;
 }
 
+static std::string base64Encode(const std::vector<unsigned char>& in) {
```
jtbandes, line 23: Any source we can link to for this implementation? Note we already have an encode implementation here: https://github.com/foxglove/ws-protocol/blob/main/cpp/examples/src/example_server.cpp#L32

foxglove_bridge_base/src/parameter.cpp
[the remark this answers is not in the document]
achim-k: Yes, I think we can improve this here. When I implemented this, I decided against doing the conversion internally but I don't remember what the reasoning was. I will see what I can do

ros2_foxglove_bridge/src/parameter_interface.cpp
achim-k: It's an adapted version from https://stackoverflow.com/a/34571089. Was not aware about our encode implementation, looks cleaner than the solution I have found

foxglove_bridge_base/src/serialization.cpp
```diff
@@ -37,6 +38,11 @@ void to_json(nlohmann::json& j, const Parameter& p) {
     j["value"] = p.getValue<double>();
   } else if (paramType == ParameterType::PARAMETER_STRING) {
     j["value"] = p.getValue<std::string>();
+  } else if (paramType == ParameterType::PARAMETER_BYTE_ARRAY) {
+    const auto paramValue = p.getValue<std::vector<unsigned char>>();
+    const std::string strValue(paramValue.begin(), paramValue.end());
+    j["value"] = base64Encode(strValue);
```
jtbandes, line 44: too bad we have to copy here, might be something to fix in a future PR?

jtbandes (approved):

foxglove_bridge_base/src/serialization.cpp
achim-k: Fixed it by using `std::string_view`

Issue #174, opened 2023-02-27
Missing state transition messages


rgov:
```
ros-noetic-foxglove-bridge/now 0.3.0-1focal.20230106.232224 arm64 [installed,local]
```

Streaming data from a system to Foxglove Studio (1.39.1; Darwin), I was surprised that my state transition panel seemed to be stuck on the same state since connection, even though the system was progressing through other states.

When I pulled the rosbag afterwards, I saw indeed the state transition messages were being sent:

<img width="737" alt="image" src="https://user-images.githubusercontent.com/108767/221616334-ca4dcd82-373f-43f7-abef-52054e7c264c.png">

(I only broadcast the message once per state transition, I don't spam the current state to the topic.)

Other messages seemed to get through; I had an image panel that updated occasionally as expected (though I didn't watch it carefully).

What is the behavior regarding when messages get dropped? If I have a high bandwidth topic, will it elbow out small topics?

achim-k:
> What is the behavior regarding when messages get dropped? If I have a high bandwidth topic, will it elbow out small topics?

The [current behavior](https://github.com/foxglove/ros-foxglove-bridge/blob/f4b2f68cde3cc87ad65cb95a36cd7690da34b3a5/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp#L1019-L1028) is rather simple and does not take into account which topic a message originates from. This is something we have to improve on in the future.

Are you sure that messages are dropped? Do you see the corresponding log messages?

On another note, the `0.3-0` release is rather outdated. I would recommend building the latest version from source or to wait for the next [noetic sync](https://discourse.ros.org/t/preparing-for-noetic-sync-2023-02-23/29929) (which seems to be just a couple of days away)

achim-k:
@rgov do you have any update on this? Do you still have this issue?

rgov:
Has anything changed with the algorithm? I haven't really used it since filing this. My system is offline for a few days.

achim-k:
> Has anything changed with the algorithm?

No it's still the same. But we are definitely open to improve on that. Have you had the same issue with `rosbridge`?

defunctzombie:
[...]

PR #201, opened 2023-03-23
Notify client when Server's send buffer limit has been reached

achim-k:
### Public-Facing Changes

- Notify client when Server's send buffer limit has been reached

### Description
Rather than only logging a message, the client is now also informed when the server is about to drop messages due to the send buffer limit being exceeded. See https://github.com/foxglove/ros-foxglove-bridge/issues/174#issuecomment-1480438800

### DIFF
--- a/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
@@ -1064,8 +1064,7 @@ inline void Server<ServerConfiguration>::sendMessage(ConnHandle clientHandle, Ch
   const auto bufferSizeinBytes = con->get_buffered_amount();
   if (bufferSizeinBytes + payloadSize >= _options.sendBufferLimitBytes) {
     const auto logFn = [this, clientHandle]() {
-      sendStatusAndLogMsg(clientHandle, StatusLevel::Warning,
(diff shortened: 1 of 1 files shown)


foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -1057,12 +1063,11 @@ inline void Server<ServerConfiguration>::sendMessage(ConnHandle clientHandle, Ch
...
   if (bufferSizeinBytes + payloadSize >= _options.sendBufferLimitBytes) {
-    FOXGLOVE_DEBOUNCE(
-      [this]() {
-        _server.get_elog().write(
-          WARNING, "Connection send buffer limit reached, messages will be dropped...");
-      },
-      2500);
+    const auto logFn = [this, clientHandle]() {
+      sendStatusAndLogMsg(clientHandle, StatusLevel::Warning,
+                          "Connection send buffer limit reached, messages will be dropped...");
```
defunctzombie, line 1068: As a user I'd find it more helpful if this said something like "Server send buffer full. Dropping message on topic /foo/bar".

achim-k: I agree that this would be more helpful, but due to log debounce, you wouldn't see it for every topic for which messages are dropped

Issue #204, opened 2023-03-30
Publishers stay alive after client channel is unadvertised


achim-k:
**Description**

After unadvertising a client channel, the publisher is still registered in the ros2 graph

- Version: https://github.com/foxglove/ros-foxglove-bridge/commit/c4ccee324c7fabb53a8b15d50241ac911c459e1c
- Platform: ubuntu 22.04 / galactic & humble

**Steps To Reproduce**
1. Start foxglove server
2. Connect with a client  (e.g. `websocat`) and send the following commands
3. `{"op":"advertise", "channels":[{"id": 1, "topic": "achim", "encoding": "cdr", "schemaName": "std_msgs/msg/String"}]}`
4. confirm with `ros2 node info /foxglove_bridge` that the publisher has been created
5. `{"op":"unadvertise", "channelIds":[1]}`
6. `ros2 node info /foxglove_bridge` still shows the publisher

**Expected Behavior**
After the channel has been advertised, `ros2 node info /foxglove_bridge` should not show the publisher anymore

achim-k:
rclcpp issue: https://github.com/ros2/rclcpp/issues/2146

defunctzombie:
@achim-k The related linear issue is closed - does that mean this issue is done?

achim-k:
Yes, forgot to close

PR #205, opened 2023-03-30
Fix publishers being created with invalid QoS profile

achim-k:
### Public-Facing Changes
- Fix publishers being created with invalid QoS profile

### Description
Since #184 we use the QoS profile of an existing publisher when creating a new publisher for a client advertised topic. In case the existing publisher has been destroyed before (but is still registered, see #204), the QoS profile is invalid and creating a publisher fails with the message `'create_publisher() failed setting topic QoS, at ./src/publisher.cpp:226'`.

This PR mitigates the issue by taking the QoS profile of an existing publisher from a different node (if there is any). The assumption is, that publishers of other nodes are less likely to be destroyed.

### DIFF
--- a/ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
+++ b/ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
@@ -669,14 +669,18 @@ private:
       const auto& topicName = advertisement.topic;
       const auto& topicType = advertisement.schemaName;
 
-      // Lookup if there are other publishers for that topic. If that's the case, we use a matching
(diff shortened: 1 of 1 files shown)


jtbandes (approved):
Seems ok, but would it be better to unregister the publisher? Or is that impossible?

achim-k:
> Seems ok, but would it be better to unregister the publisher? Or is that impossible?

That would be ideal, but there seems to be a bug which prevents us from doing so. See https://github.com/foxglove/ros-foxglove-bridge/issues/204 and the linked rclcpp issue

PR #210, opened 2023-04-10
Catch exceptions thrown in handler functions, send status to client

### LINKED ISSUE #148: Handler functions do not give feedback if operation succeeded/failed
**Description**

We currently have the following handler functions which all return `void`, leaving the server clueless if the operation succeeded or not (unless the function throws). We should consider changing the function signatures to return a bool or something similar, such that the server can send a message to the client if the operation failed.

https://github.com/foxglove/ros-foxglove-bridge/blob/d29159c10127ecfac9262ddaa657def371f1aeab/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp#L140-L149

achim-k:
### Public-Facing Changes
None

### Description
- Refactors the way handler functions are run in a separate thread. It's better to do that in a unified way in the server rather than doing that for ROS 1 & 2 differently as it was implemented in #165
- Catch exeptions thrown in handler functions and send status message to client

Fixes #148

### DIFF
--- a/foxglove_bridge_base/include/foxglove_bridge/callback_queue.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/callback_queue.hpp
@@ -8,14 +8,15 @@
 #include <thread>
 #include <vector>
 
-#include <rclcpp/logging.hpp>
+#include "websocket_logging.hpp"
 
-namespace foxglove_bridge {
+namespace foxglove {
 
 class CallbackQueue {
 public:
-  CallbackQueue(size_t numThreads = 1)
-      : _quit(false) {
+  CallbackQueue(LogCallback logCallback, size_t numThreads = 1)
+      : _logCallback(logCallback)
+      , _quit(false) {
     for (size_t i = 0; i < numThreads; ++i) {
       _workerThreads.push_back(std::thread(&CallbackQueue::doWork, this));
     }
@@ -59,13 +60,15 @@ private:
           cb();
         } catch (const std::exception& ex) {
           // Should never get here if we catch all exceptions in the callbacks.
-          RCLCPP_ERROR(rclcpp::get_logger("foxglove_bridge"),
-                       "Caught unhandled exception in calback_queue: %s", ex.what());
+          const std::string msg =
+            std::string("Caught unhandled exception in calback_queue") + ex.what();
+          _logCallback(WebSocketLogLevel::Error, msg.c_str());
         }
       }
     }
   }
 
+  LogCallback _logCallback;
   std::atomic<bool> _quit;
   std::mutex _mutex;
   std::condition_variable _cv;
@@ -73,4 +76,4 @@ private:
   std::vector<std::thread> _workerThreads;
 };
 
-}  // namespace foxglove_bridge
+}  // namespace foxglove
--- a/foxglove_bridge_base/include/foxglove_bridge/callback_queue.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/callback_queue.hpp
@@ -63,6 +63,8 @@ private:
           const std::string msg =
             std::string("Caught unhandled exception in calback_queue") + ex.what();
           _logCallback(WebSocketLogLevel::Error, msg.c_str());
+        } catch (...) {
+          _logCallback(WebSocketLogLevel::Error, "Caught unhandled exception in calback_queue");
         }
       }
     }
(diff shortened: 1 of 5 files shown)


foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -227,6 +230,9 @@ inline Server<ServerConfiguration>::Server(std::string name, LogCallback logger,
     std::bind(&Server::handleMessage, this, std::placeholders::_1, std::placeholders::_2));
   _server.set_reuse_addr(true);
   _server.set_listen_backlog(128);
+
+  // Callback queue for handling client requests.
+  _handlerCallbackQueue = std::make_unique<CallbackQueue>(_logger, 1ul /* 1 thread */);
```
jtbandes, line 235: I've seen it done this way which I like since it uses the same parameter name.
```suggestion
  _handlerCallbackQueue = std::make_unique<CallbackQueue>(_logger, /*numThreads=*/ 1ul);
```

jtbandes (approved):
This seems like a good change overall. I have seen C++ codebases benefit greatly from a Status type, e.g. https://abseil.io/docs/cpp/guides/status, so that would be a nice improvement/cleanup to consider. I'm not sure if there is one already in the ROS package ecosystem?

achim-k:
> This seems like a good change overall. I have seen C++ codebases benefit greatly from a Status type, e.g. https://abseil.io/docs/cpp/guides/status, so that would be a nice improvement/cleanup to consider. I'm not sure if there is one already in the ROS package ecosystem?

I wonder if we should not simply throw exceptions if something went wrong. We already use exceptions in other places, so there is no good reason to not use them. This seems to be also what the C++ core guidelines recommend: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e25-if-you-cant-throw-exceptions-simulate-raii-for-resource-management

foxglove_bridge_base/include/foxglove_bridge/server_interface.hpp
```diff
@@ -17,6 +17,63 @@ constexpr size_t DEFAULT_SEND_BUFFER_LIMIT_BYTES = 10000000UL;  // 10 MB
...
+};
+class ServiceUnknownError : public ServiceError {
+  using ServiceError::ServiceError;
+};
+class ServiceUnavailableError : public ServiceError {
+  using ServiceError::ServiceError;
+};
+class ServiceSchemaError : public ServiceError {
+  using ServiceError::ServiceError;
+};
```
jtbandes, line 75: Since we already have a detailed message as part of the error, I wonder if we really need a separate subclass for each type of error, or could you just use ServiceError for all of these?

jtbandes (approved):

foxglove_bridge_base/include/foxglove_bridge/server_interface.hpp
achim-k: This is part of the public API, so not sure what error granularity is best here. I guess I will remove all error types that are not explicitly caught in the server code

foxglove_bridge_base/include/foxglove_bridge/callback_queue.hpp
```diff
@@ -59,18 +60,20 @@ class CallbackQueue {
           cb();
         } catch (const std::exception& ex) {
           // Should never get here if we catch all exceptions in the callbacks.
-          RCLCPP_ERROR(rclcpp::get_logger("foxglove_bridge"),
-                       "Caught unhandled exception in calback_queue: %s", ex.what());
+          const std::string msg =
+            std::string("Caught unhandled exception in calback_queue") + ex.what();
+          _logCallback(WebSocketLogLevel::Error, msg.c_str());
```
achim-k, line 65: Add a `catch(...)` clause here

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -547,11 +552,22 @@ inline void Server<ServerConfiguration>::handleMessage(ConnHandle hdl, MessagePt
...
       } break;
       case OpCode::BINARY: {
-        const auto& payload = msg->get_payload();
-        handleBinaryMessage(hdl, reinterpret_cast<const uint8_t*>(payload.data()), payload.size());
+        _handlerCallbackQueue->addCallback([this, hdl, msg]() {
+          try {
+            handleBinaryMessage(hdl, msg);
+          } catch (const std::exception& e) {
+            sendStatusAndLogMsg(hdl, StatusLevel::Error, e.what());
+          }
```
achim-k, line 569: Add a `catch(...)` clause here

PR #211, opened 2023-04-10
Add parameter for minimum subscription QoS depth

### LINKED ISSUE #208: Subscribers always using qos depth of 1 with fastDDS
**Description**
Subscribers created by the bridge always using qos depth of 1 with fastDDS

- Version: ca7d0c30910a97ccc56d00f25371d21e6da21eab
- Platform: Ubuntu 20.04

**Steps To Reproduce**
Currently, get_publishers_info_by_topic will report all publishers as having a qos depth of 0 with fastDDS. I opened a ticket for that here https://github.com/ros2/rmw_fastrtps/issues/684
This leads to foxglove_bridge using a qos depth of 1 for all subscribers when using fastDDS - which causes lots of dropped messages (a real headache for /tf visualization specifically). Hopefully, that issue with fastDDS will be resolved, but in the interim I suggest increasing the minimum qos depth for subscribers to alleviate this issue (internally, we're going to be patching foxglove_bridge for the time being to increase the minimum depth to 10).

**Expected Behavior**
Expect foxglove_bridge to use reasonable qos depth on subscribers when using fastDDS.

achim-k:
### Public-Facing Changes
- Add parameter for minimum subscription QoS depth [ROS2]

### Description
Adds an additional parameter for specifying the minimum QoS depth for subscriptions. Fixes #208

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Parameters are provided to configure the behavior of the bridge. These parameter
  * __use_compression__: Use websocket compression (permessage-deflate). Suited for connections with smaller bandwith, at the cost of additional CPU load.
  * (ROS 1) __max_update_ms__: The maximum number of milliseconds to wait in between polling `roscore` for new topics, services, or parameters. Defaults to `5000`.
  * (ROS 2) __num_threads__: The number of threads to use for the ROS node executor. This controls the number of subscriptions that can be processed in parallel. 0 means one thread per CPU core. Defaults to `0`.
- * (ROS 2) __min_qos_depth__: Minimum depth used for the QoS profile of subscriptions. Defaults to `1`.
+ * (ROS 2) __min_qos_depth__: Minimum depth used for the QoS profile of subscriptions. Defaults to `1`. This is to set a lower limit for a subscriber's QoS depth which is computed by summing up depths of all publishers. See also [#208](https://github.com/foxglove/ros-foxglove-bridge/issues/208).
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -78,6 +78,7 @@ Parameters are provided to configure the behavior of the bridge. These parameter
  * __use_compression__: Use websocket compression (permessage-deflate). Suited for connections with smaller bandwith, at the cost of additional CPU load.
  * (ROS 1) __max_update_ms__: The maximum number of milliseconds to wait in between polling `roscore` for new topics, services, or parameters. Defaults to `5000`.
  * (ROS 2) __num_threads__: The number of threads to use for the ROS node executor. This controls the number of subscriptions that can be processed in parallel. 0 means one thread per CPU core. Defaults to `0`.
+ * (ROS 2) __min_qos_depth__: Minimum depth used for the QoS profile of subscriptions. Defaults to `1`.
```
jtbandes, line 81: Maybe explain more about why the min is configurable? What will the qos depth be if it's not the min value?

jtbandes (approved):

README.md
achim-k:
```suggestion
 * (ROS 2) __min_qos_depth__: Minimum depth used for the QoS profile of subscriptions. Defaults to `1`. This is to set a lower limit for a subscriber's QoS depth which is computed by summing up depths of all publishers.
```

PR #214, opened 2023-04-13
Add support for action topic and services (ROS2)

achim-k:
### Public-Facing Changes
- Add support for action topic and services (ROS2)

### Description
Adds support for parsing .action files to retrieve message definitions for the following action subtypes:
- feedback topic (`_FeedbackMessage`) *
- SendGoal
  - request (`_SendGoal_Request`) *
  - response (`_SendGoal_Response`)
- GetResult
  - request (`_GetResult_Request`)
  - response (`_GetResult_Response`) *

Unfortunately, the action definition in the .action file does not include other fields such as the goal_id which are sent as part of the feedback message or SendGoal_Request and GetResult_Response. These types are therefore added to the message definition such that the entire schema is generated correctly.

Parsing IDL action definitions is currently not supported.

Example usage in Studio:
image

### DIFF
--- a/ros2_foxglove_bridge/src/message_definition_cache.cpp
+++ b/ros2_foxglove_bridge/src/message_definition_cache.cpp
@@ -123,7 +123,7 @@ static std::vector<std::string> split_string(const std::string& str,
 /// @return A tuple holding goal, result and feedback definitions
 static std::tuple<std::string, std::string, std::string> split_action_msg_definition(
   const std::string& action_definition) {
-  constexpr char SEP[] = "---";
+  constexpr char SEP[] = "\n---\n";
 
   const auto definitions = split_string(action_definition, SEP);
   if (definitions.size() != 3) {
--- a/ros2_foxglove_bridge/src/message_definition_cache.cpp
+++ b/ros2_foxglove_bridge/src/message_definition_cache.cpp
@@ -14,38 +14,11 @@
 #include <ament_index_cpp/get_resources.hpp>
 #include <rcutils/logging_macros.h>
 
-namespace {
-
-/// @brief Split an action definition into individual goal, result and feedback definitions.
-/// @param action_definition The full action definition as read from a .action file
-/// @return A tuple holding goal, result and feedback definitions
-std::tuple<std::string, std::string, std::string> split_action_msg_definition(
-  const std::string& action_definition) {
-  constexpr char SEP[] = "---";
-  constexpr size_t SEP_SIZE = sizeof(SEP);
-  const auto first_sep_idx = action_definition.find(SEP);
-  const auto last_sep_idx = action_definition.find(SEP, first_sep_idx + SEP_SIZE);
-  const size_t size_result_def = last_sep_idx - first_sep_idx;
-
-  if (first_sep_idx == std::string::npos || last_sep_idx == std::string::npos ||
-      first_sep_idx == last_sep_idx) {
-    throw std::invalid_argument("Invalid action definition:\n" + action_definition);
-  }
-
-  const std::string goal_def = action_definition.substr(0UL, first_sep_idx);
-  const std::string result_def =
-    action_definition.substr(first_sep_idx + SEP_SIZE, size_result_def - SEP_SIZE);
-  const std::string feedback_def = action_definition.substr(last_sep_idx + SEP_SIZE);
-  return {goal_def, result_def, feedback_def};
-}
-
-}  // namespace
-
 namespace foxglove {
 
 // Match datatype names (foo_msgs/Bar or foo_msgs/msg/Bar or foo_msgs/srv/Bar)
 static const std::regex PACKAGE_TYPENAME_REGEX{
-  R"(^([a-zA-Z0-9_]+)/((?:msg|srv|action)?)/?([a-zA-Z0-9_]+)$)"};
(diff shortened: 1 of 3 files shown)


ros2_foxglove_bridge/src/message_definition_cache.cpp
```diff
@@ -14,11 +14,38 @@
...
+}
+
+}  // namespace
+
 namespace foxglove {
 
 // Match datatype names (foo_msgs/Bar or foo_msgs/msg/Bar or foo_msgs/srv/Bar)
 static const std::regex PACKAGE_TYPENAME_REGEX{
-  R"(^([a-zA-Z0-9_]+)/(?:msg/|srv/|action/)?([a-zA-Z0-9_]+)$)"};
+  R"(^([a-zA-Z0-9_]+)/((?:msg|srv|action)?)/?([a-zA-Z0-9_]+)$)"};
```
jtbandes, line 48: the `/` after the subtype name should not be optional

ros2_foxglove_bridge/src/message_definition_cache.cpp
```diff
@@ -166,15 +202,53 @@ const MessageSpec& MessageDefinitionCache::load_message_spec(
...
-  return spec;
+      // Define type definitions for each action subtype.
+      // These type definitions may include additional fields such as the goal_id.
+      // See also https://design.ros2.org/articles/actions.html
+      const std::map<std::string, std::string> action_type_definitions = {
+        {"_FeedbackMessage", "unique_identifier_msgs/UUID goal_id\n" + feedbackDef},
+        {"_GetResult_Request", "unique_identifier_msgs/UUID goal_id\n"},
+        {"_GetResult_Response", "int8 status\n" + resultDef},
+        {"_SendGoal_Request", "unique_identifier_msgs/UUID goal_id\n" + goalDef},
+        {"_SendGoal_Response", "bool accepted\nbuiltin_interfaces/msg/Time stamp"}};
```
jtbandes, line 217: The linked design doc discusses these fields at a high level, but where is it documented that these are actually added to the definitions?

ros2_foxglove_bridge/src/message_definition_cache.cpp
```diff
@@ -166,15 +202,53 @@ const MessageSpec& MessageDefinitionCache::load_message_spec(
...
+
+      // Create a MessageSpec instance for every action subtype and add it to the cache.
+      for (const auto& [action_suffix, definition] : action_type_definitions) {
+        DefinitionIdentifier definitionIdentifier;
```
jtbandes, line 221: I see mixed snake & camel case in this file, can we standardize on one?

ros2_foxglove_bridge/src/message_definition_cache.cpp
```diff
@@ -137,7 +164,16 @@ const MessageSpec& MessageDefinitionCache::load_message_spec(
...
   }
   const std::string package = match[1].str();
-  const std::string filename = match[2].str() + extension_for_format(definition_identifier.format);
+  const std::string subfolder = match[2].str();
+  const std::string type_name = match[3].str();
+  const bool is_action_type = subfolder == "action";
+
+  // The action type name includes the subtype which we have to remove to get the action name.
+  // Type name: Fibonacci_FeedbackMessage -> Action name: Fibonacci
+  const std::string action_name = type_name.substr(0UL, type_name.find('_'));
```
jtbandes, line 173: do we need to handle type names that also contain underscores? do we need to verify that the suffix we stripped off is actually one of the suffixes we support?

ros2_foxglove_bridge/src/message_definition_cache.cpp
achim-k: I also have not found proper documentation for that. Best documentation I could find is the source code that generates the action types:

https://github.com/ros2/rosidl/blob/a57baea5dee6108246db9991e39d28a10a7ee2b0/rosidl_parser/rosidl_parser/definition.py#L720-L727

ros2_foxglove_bridge/src/message_definition_cache.cpp
achim-k: Yes, we should. Will do that

ros2_foxglove_bridge/src/message_definition_cache.cpp
```diff
@@ -145,6 +118,44 @@ static std::vector<std::string> split_string(const std::string& str,
...
 }
 
+/// @brief Split an action definition into individual goal, result and feedback definitions.
+/// @param action_definition The full action definition as read from a .action file
+/// @return A tuple holding goal, result and feedback definitions
+static std::tuple<std::string, std::string, std::string> split_action_msg_definition(
+  const std::string& action_definition) {
+  constexpr char SEP[] = "---";
+
+  const auto definitions = split_string(action_definition, SEP);
```
jtbandes, line 128: What if one of the segments has a string constant with `---` in it? e.g. `string FOO=---` ?

jtbandes (approved):
Seems alright. I do think we should put this behind an explicit cli flag because it's a niche/advanced/experimental use case for which we don't provide a great UX in studio.

achim-k:
> Seems alright. I do think we should put this behind an explicit cli flag because it's a niche/advanced/experimental use case for which we don't provide a great UX in studio.

I have submitted https://github.com/foxglove/ros-foxglove-bridge/pull/216 for that purpose. Separate PR, since certain hidden action topics (e.g. `.../_action/status`) already got advertised before this PR

PR #216, opened 2023-04-14
Add parameter to include hidden topics and services (ROS 2)

achim-k:
### Public-Facing Changes

- Add parameter to include hidden topics and services (not included by default)

### Description
Adds the new parameter `include_hidden` which controls if hidden topics or services are advertised.

From https://design.ros2.org/articles/topic_and_service_names.html#hidden-topic-or-service-names:

> Any topic or service name that contains any tokens (either namespaces or a topic or service name) that start with an underscore (_) will be considered hidden and tools may not show them unless explicitly asked.


jtbandes (approved):
🎉

Issue #217, opened 2023-04-20
'Unsupported parameter array type' breaks foxglove bridge


ysuurmeijer:
Receiving the following breaking error when trying to launch foxglove_bridge on a ROS network with a parameter of type array. The error results in shutdown of the nodelet manager. 

`[ERROR] [1682013405.757034088]: Invalid parameter: Parameter '/some_node/inputs': Unsupported parameter array type
terminate called after throwing an instance of 'XmlRpc::XmlRpcException'`

Temporary workaround is to disable the parameterSubscribe capability `roslaunch foxglove_bridge foxglove_bridge.launch capabilities:=[clientPublish,services,connectionGraph]`

- Version: 0.5.3
- Platform: Linux / AMD64 / ROS Noetic

Launch `foxglove-bridge` on a ROS network with and upload the following parameter file under the `some_node` namespace
```
inputs:
  - topic: /gps/fix
    message_type: NavSatFix
  - topic: /gps2/fix
    message_type: NavSatFix
  - topic: /gps3/fix
    message_type: NavSatFix
```

achim-k:
> The error results in shutdown of the nodelet manager.

Are you sure about that? It should just log an error message in that case

achim-k:
Crash has been fixed in #218 
Closing this issue in favor of #219

PR #218, opened 2023-04-20
Fix unhandled xmlrpc exception

achim-k:
### Public-Facing Changes
- Fix unhandled xmlrpc exception causing nodelet to crash (ROS1)

### Description
Addresses the crash reported in #217

### DIFF
--- a/ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
+++ b/ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
@@ -708,7 +708,11 @@ private:
         getMTNodeHandle().getParam(paramName, value);
         params.push_back(fromRosParam(paramName, value));
       } catch (const std::exception& ex) {
-        ROS_ERROR("Invalid parameter: %s", ex.what());
+        ROS_ERROR("Invalid parameter '%s': %s", paramName.c_str(), ex.what());
+      } catch (const XmlRpc::XmlRpcException& ex) {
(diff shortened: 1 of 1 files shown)


jtbandes (approved):

ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
```diff
@@ -805,8 +809,11 @@ class FoxgloveBridge : public nodelet::Nodelet {
       const auto param = fromRosParam(paramName, paramValue);
       _server->updateParameterValues({param});
     } catch (const std::exception& ex) {
-      ROS_ERROR("Failed to convert parameter: %s", ex.what());
-      return;
+      ROS_ERROR("Failed to update parameter: %s", ex.what());
+    } catch (const XmlRpc::XmlRpcException& ex) {
+      ROS_ERROR("Failed to update parameter: %s", ex.getMessage().c_str());
+    } catch (...) {
+      ROS_ERROR("Failed to update parameter");
```
jtbandes, line 816: Is there any way to include more info about the exception type in the log? I googled a little bit and couldn't find anything... I see that you can use `current_exception()` to get an `exception_ptr`, but then there's nothing useful you can do with that 😬

achim-k: I'm also not aware that this would be possible. Also just learned that you can literally throw anything, e.g. just `throw 1;`

PR #221, opened 2023-05-01
Add support for nested parameters (ROS1)

### LINKED ISSUE #219: [ROS1] Arrays of parameter maps are not supported
**Description**
See also #217 

```
inputs:
  - topic: /gps/fix
    message_type: NavSatFix
  - topic: /gps2/fix
    message_type: NavSatFix
  - topic: /gps3/fix
    message_type: NavSatFix
```

If the above parameter `inputs` is present, foxglove bridge logs the following error: `Unsupported parameter array type`

**Steps To Reproduce**
- Copy the above yaml into `params.yaml`
- `rosparam load params.yaml`
- Launch foxglove_bridge and connect with Foxglove Studio

**Expected Behavior**
All parameter types should be supported

achim-k:
### Public-Facing Changes

- Add support for nested parameters (ROS1)

### Description
This PR adds support for parameter maps and arrays of parameter maps (ROS 1 only). To accomplish that, the existing `Parameter` class has been split up into a `ParameterValue` class which facilitates conversion to/from `xmlrpc::XmlRpcValue`.

Spec update: https://github.com/foxglove/ws-protocol/pull/434

Fixes #219

### DIFF
--- a/foxglove_bridge_base/include/foxglove_bridge/parameter.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/parameter.hpp
@@ -3,6 +3,7 @@
 #include <any>
 #include <stdint.h>
 #include <string>
+#include <unordered_map>
 #include <vector>
 
 namespace foxglove {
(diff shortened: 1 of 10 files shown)


jtbandes (commented):
Do we have a corresponding spec change for this?

achim-k:
> Do we have a corresponding spec change for this?

You are right, completely forgot that since it already works with Studio out of the box. Will update the spec

jtbandes (approved):

PR #228, opened 2023-05-17
Extract ROS 2 bridge header

mvukov:
### Public-Facing Changes

Extracts ROS 2 bridge header such that the node class can be used directly, not only as a component.

### Description

I want to compile and use the bridge in https://github.com/mvukov/rules_ros2. I don't have support for components, yet, so I'd like to use the bridge library itself. For that reason, I extracted the header file.

There are no functional changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -244,7 +244,7 @@ if(ROS_BUILD_TYPE STREQUAL "catkin")
       DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/launch
     )
 elseif(ROS_BUILD_TYPE STREQUAL "ament_cmake")
-    install(FILES ros2_foxglove_bridge/include/foxglove_bridge/foxglove_bridge.hpp
+    install(FILES ros2_foxglove_bridge/include/foxglove_bridge/ros2_foxglove_bridge.hpp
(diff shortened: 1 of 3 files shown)


mvukov:
> I think we can rename the header from ros2_foxglove_bridge.hpp -> foxglove_bridge.hpp

Should I also rename the .cpp file for consistency?

> Could you change ros2_foxglove_bridge_node.cpp to use that new header?

AFAICS, the node doesn't need the header as it loads the component in runtime. WDYT?

achim-k:
> Should I also rename the .cpp file for consistency?

Yes, good idea

> AFAICS, the node doesn't need the header as it loads the component in runtime. WDYT?

Not super important actually. The node starts an extra component manager node which we could avoid when launching the foxglove bridge node directly (using the header). But this is something I can always do later, so indeed no need to do this now.

mvukov:
In fact, I can't rename the header to foxglove_bridge.hpp because there already exists one in the base library. The import path `foxglove_bridge/foxglove_bridge.hpp` is then ambiguous. Please let me know how you'd like to proceed.

mvukov:
> You will have to add a install rule for that public header (see below)

Done. @achim-k PTAL.

PR #230, opened 2023-05-28
Fix websocketpp include problem

tudoroancea:
### Public-Facing Changes

None

### Description

I encountered problems when building `ros-foxglove-bridge` from source on macOS and Linux that were actually include problems linked to `websocketpp` (as described [in this discussion](https://github.com/orgs/foxglove/discussions/624)).
I simply added an automatic command in the `CMakeLists.txt` that finds the include directory of `websocketpp` after the command `find_package(websocketpp REQUIRED)` succeeds, and then adds it to the include dirs of the `foxglove_bridge_base`.


tudoroancea:
Small caveat: on Linux, the header files in `/usr/include` are not included properly

achim-k:
To make the MacOS build work for robostack, we applied the following patch:

https://github.com/defunctzombie/ros-humble/blob/9522544525af3f9dd8be3273832d9c4883067826/patch/ros-humble-foxglove-bridge.patch

Could you verify if that works for you?

CMakeLists.txt
```diff
@@ -70,7 +72,8 @@ target_include_directories(foxglove_bridge_base
   PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/foxglove_bridge_base/include>
     $<INSTALL_INTERFACE:include>
-)
+  )
+target_include_directories(foxglove_bridge_base SYSTEM PUBLIC ${WEBSOCKETPP_INCLUDE_DIR} /usr/local/include)
```
achim-k, line 76: Including `/usr/local/include` will not work under Windows

CMakeLists.txt
```diff
@@ -18,16 +18,18 @@ macro(enable_strict_compiler_warnings target)
   if (MSVC)
     target_compile_options(${target} PRIVATE /WX /W4)
   elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-    target_compile_options(${target} PRIVATE -Wall -Wextra -Wpedantic -Werror -Wold-style-cast -Wfloat-equal -Wmost -Wunused-exception-parameter)
+    target_compile_options(${target} PRIVATE -Wall -Wextra -Wpedantic -Werror -Wfloat-equal -Wmost -Wunused-exception-parameter)
```
achim-k, line 21: Why do you remove `-Wold-style-cast` here? Because it causes warnings in header files stemming from websocketpp/asio? I think the better approach is to make these header files being treated as system includes

achim-k (changes requested):

defunctzombie:
Seems this PR has gone stale? Is this still a problem or can we close this and revisit if there's interest again?

jtbandes:
Closing for now but we can reopen if there's further interest.

PR #231, opened 2023-05-28
Factor out getRosDistro

mvukov:
### Public-Facing Changes

None.

### Description

Robustified fetching of ROS_DISTRO env var.

### DIFF
--- a/ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
+++ b/ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
@@ -10,6 +10,15 @@ inline bool isHiddenTopicOrService(const std::string& name) {
   }
   return name.front() == '_' || name.find("/_") != std::string::npos;
 }
+
+std::string getRosDistro() {
+  const char* rosDistro = std::getenv("ROS_DISTRO");
+  if (rosDistro == nullptr) {
+    return "unknown";
+  }
+  return rosDistro;
+}
+
 }  // namespace
 
 using namespace std::chrono_literals;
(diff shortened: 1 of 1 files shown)


mvukov:
@achim-k PTAL.

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -10,6 +10,15 @@ inline bool isHiddenTopicOrService(const std::string& name) {
   }
   return name.front() == '_' || name.find("/_") != std::string::npos;
 }
+
+std::string getRosDistro() {
+  const char* rosDistro = std::getenv("ROS_DISTRO");
+  if (rosDistro == nullptr) {
+    return "unknown";
```
achim-k, line 17: I think it would be good to log a warning if that happens

achim-k (commented):
LGTM. Could you apply the changes also for the ROS1 node? Thanks!

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
mvukov: OK, will try to make that happen. Do the bridge and/or studio rely on this env var?

achim-k: Yes, Studio uses it to [determine the profile and to load known datatypes](https://github.com/foxglove/studio/blob/79b42ca3e7ad20cc225b094dd791d9e273edf14f/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts#L285-L304). This is e.g. necessary for publishing messages from Studio

defunctzombie: Also worth considering is a fatal error here. This variable should be set for good operation.

defunctzombie:
Closing this due to lack of activity from the last feedback. We are happy to re-visit if you pick this up at a later time.

PR #232, opened 2023-06-14
Implement ws-protocol's `fetchAsset` specification

achim-k:
### Public-Facing Changes

- Implements https://github.com/foxglove/ws-protocol/pull/462 to allow fetching of assets

### Description
Implements the above mentioned spec to fetch assets which will allow Foxglove Studio to request assets (such as URDF mesh files) over the websocket connection.


ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -791,6 +797,28 @@ void FoxgloveBridge::serviceRequest(const foxglove::ServiceRequest& request,
...
+void FoxgloveBridge::fetchAsset(const std::string& assetId, uint32_t requestId,
+                                ConnectionHandle clientHandle) {
+  resource_retriever::Retriever retriever;
+  foxglove::FetchAssetResponse response;
+  response.requestId = requestId;
+
+  try {
+    const resource_retriever::MemoryResource memoryResource = retriever.get(assetId);
+    response.status = foxglove::FetchAssetStatus::Success;
+    response.mediaTypeOrMsg = "";
```
achim-k, line 809: TODO: Defer mediaType from asset name

README.md
```diff
@@ -76,7 +76,8 @@ Parameters are provided to configure the behavior of the bridge. These parameter
...
  * __send_buffer_limit__: Connection send buffer limit in bytes. Messages will be dropped when a connection's send buffer reaches this limit to avoid a queue of outdated messages building up. Defaults to `10000000` (10 MB).
  * __use_compression__: Use websocket compression (permessage-deflate). Suited for connections with smaller bandwith, at the cost of additional CPU load.
- * __capabilities__: List of supported [server capabilities](https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md). Defaults to `[clientPublish,parameters,parametersSubscribe,services,connectionGraph]`.
+ * __capabilities__: List of supported [server capabilities](https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md). Defaults to `[clientPublish,parameters,parametersSubscribe,services,connectionGraph,assets]`.
+ * __asset_uri_whitelist__: List of regular expressions ([ECMAScript grammar](https://en.cppreference.com/w/cpp/regex/ecmascript)) of allowed asset URIs. Defaults to `["package://.*"]`.
```
jtbandes, line 80: How does a user know what patterns would be valuable to put in here? There is no documentation explaining how `file://` urls work.

ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
```diff
@@ -844,6 +859,29 @@ class FoxgloveBridge : public nodelet::Nodelet {
...
+  void fetchAsset(const std::string& uri, uint32_t requestId, ConnectionHandle clientHandle) {
+    foxglove::FetchAssetResponse response;
+    response.requestId = requestId;
+
+    try {
+      if (!isWhitelisted(uri, _assetUriWhitelistPatterns)) {
+        throw std::runtime_error("Asset URI not allowed: " + uri);
+      }
+
+      const resource_retriever::MemoryResource memoryResource = _resource_retriever.get(uri);
```
jtbandes, line 871: Is there any way to limit the ability to access local files within a certain directory? It seems like a big security hole if this can access any file on the local machine.

README.md
jtbandes: `allowlist`?

https://inclusivenaming.org/word-lists/tier-1/whitelist/

ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
achim-k: > Is there any way to limit the ability to access local files within a certain directory? It seems like a big security hole if this can access any file on the local machine.

Agreed, that's why I added the allowlist parameter. If you only want to specify certain package folders, you would have to list them manually.  `file://` URIs are by default not allowed.

Example
- `['package://robot_description/.+']` - allow only access to `robot_description` package folder
- `['package://.+[.]dae$','package://.+[.]stl$']` - only allow fetching of `dae` & `stl` files from package folders

Maybe the default `['package://.*']` is too permissive. Should I change it to only allow urdf, dae & stl files?

README.md
achim-k: It uses [resource_retriever](https://index.ros.org/p/resource_retriever/github-ros-resource_retriever) under the hood which is able to resolve `package://` URLs but also `file://` and `http(s)://`. Will extend the documentation

ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
achim-k: Bummer: with the default `['package://.*']` one can specify relative paths to retrieve files that are not intended to be fetched as assets: 

`"uri": "package://ur_description/../../../secret.txt"`

jtbandes: Yep `..` is what I was thinking of.
If there is some upstream infrastructure we depend on then maybe there's not a lot we can do about this, but we should at least provide people with a warning before they enable this feature.

achim-k: > If there is some upstream infrastructure we depend on then maybe there's not a lot we can do about this, but we should at least provide people with a warning in documentation before they enable this feature.

The default regex is now a lot stricter. Do you think that's sufficient or should we the asset feature by default?

foxglove_bridge_base/include/foxglove_bridge/common.hpp
```diff
@@ -145,4 +147,16 @@ struct ServiceResponse {
 
 using ServiceRequest = ServiceResponse;
 
+enum FetchAssetStatus {
```
jtbandes, line 150: - Prefer `enum class` to avoid polluting the namespace
- Add `: uint8_t` to match the spec

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -906,6 +909,22 @@ inline void Server<ServerConfiguration>::handleTextMessage(ConnHandle hdl, Messa
                             "Client was not subscribed to connection graph updates");
       }
     } break;
+    case FETCH_ASSET: {
+      if (!_handlers.fetchAssetHandler) {
+        return;
+      }
+
+      const auto uri = payload.at("uri").get<std::string>();
+      const auto requestId = payload.at("requestId").get<uint32_t>();
```
jtbandes, line 918: what happens if these are missing or the wrong type? if an exception is thrown here, will it be caught somewhere?

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -1371,4 +1390,36 @@ inline bool Server<ServerConfiguration>::hasCapability(const std::string& capabi
...
+template <typename ServerConfiguration>
+inline void Server<ServerConfiguration>::sendFetchAssetResponse(
+  ConnHandle clientHandle, const FetchAssetResponse& response) {
+  std::error_code ec;
+  const auto con = _server.get_con_from_hdl(clientHandle, ec);
+  if (ec || !con) {
+    return;
+  }
+
+  const size_t dataSize = response.status == FetchAssetStatus::Success ? response.data.size() : 0ul;
```
jtbandes, line 1402: do we need to similarly check if status==Error before adding `errorMessage.size()` to the `messageSize`?

foxglove_bridge_base/src/test/test_client.cpp
```diff
@@ -132,7 +132,35 @@ std::future<Channel> waitForChannel(std::shared_ptr<ClientInterface> client,
         }
       }
     });
+  return future;
+}
+
+std::future<FetchAssetResponse> waitForFetchAssetResponse(std::shared_ptr<ClientInterface> client) {
+  auto promise = std::make_shared<std::promise<FetchAssetResponse>>();
```
jtbandes, line 139: nit: is make_shared required?

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -814,6 +820,35 @@ void FoxgloveBridge::serviceRequest(const foxglove::ServiceRequest& request,
...
+    // We reject URIs that are not on the allowlist or that contain two consecutive dots. The latter
+    // can be utilized to construct URIs for retrieving confidential files that should not be
+    // accessible over the WebSocket connection. Example:
+    // `package://<pkg_name>/../../../secret.txt`. This is an extra security measure and should not
+    // be necessary if the allowlist is strict enough.
+    if (uri.find("..") != std::string::npos || !isWhitelisted(uri, _assetUriAllowlistPatterns)) {
+      throw std::runtime_error("Asset URI not allowed: " + uri);
+    }
+
+    const resource_retriever::MemoryResource memoryResource = _resource_retriever.get(uri);
```
jtbandes, line 838: so this will block, even for HTTP fetches? Is there any built-in way to do this in a non-blocking fashion or should we consider doing it in a thread (if the API is thread safe)?

jtbandes (approved):

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
achim-k: Yes, any exception that is not handled in the operation handler would be caught here:
https://github.com/foxglove/ros-foxglove-bridge/blob/9a206a06a2e8fd1564f066840a2b7e86d2cf74b7/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp#L560-L571

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
achim-k: You are right, this will block and prevent any other client operation from being handled since we use [only a single thread for handling client requests](https://github.com/foxglove/ros-foxglove-bridge/blob/9a206a06a2e8fd1564f066840a2b7e86d2cf74b7/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp#L239-L240). I guess that this wouldn't be much of a problem for file:// or package:// URIs as these will succeed/fail fairly quickly, but for http it might indeed be problematic. I will see what I can do

foxglove_bridge_base/src/test/test_client.cpp
achim-k: You left a similar comment here: https://github.com/foxglove/ros-foxglove-bridge/pull/239#discussion_r1253489103

Will create an issue to keep track of this

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -846,7 +856,9 @@ void FoxgloveBridge::fetchAsset(const std::string& uri, uint32_t requestId,
     response.errorMessage = "Failed to retrieve asset " + uri;
   }
 
-  _server->sendFetchAssetResponse(clientHandle, response);
+  if (_server) {
+    _server->sendFetchAssetResponse(clientHandle, response);
```
jtbandes, line 860: Is this method thread safe?

achim-k: Yes. It's basically the same as `_server->sendMessage` which is also called by multiple threads (if subscribed to multiple topics)

Issue #200, opened 2023-03-17
create_publisher() failed setting topic QoS, at ./src/publisher.cpp:226'


sgf201:
Description
create_publisher() failed setting topic QoS, at ./src/publisher.cpp:226'
the topic is not advertised

    Version:ROS2 humble bridge with WebSocket++/0.8.2
    Platform: ubuntu22

Steps To Reproduce

Qos set to rclcpp::QoS qos(rclcpp::KeepLast{7});

Expected Behavior
topic advertised

            [rcutils|error_handling.c:108] rcutils_set_error_state()
            This error state is being overwritten:

'Unknown QoS history policy, at ./src/qos.cpp:61'

with this new error message:

'create_publisher() failed setting topic QoS, at ./src/publisher.cpp:226'

rcutils_reset_error() should be called after error handling to avoid this.
<<<
[ERROR] [1678937020.102152168] [foxglove_bridge]: Failed to create publisher: could not create publisher: create_publisher() failed setting topic QoS, at ./src/publisher.cpp:226, at ./src/rcl/publisher.c:116
[WARN] [1678937020.102373126] [foxglove_bridge]: Ignoring client unadvertisement from 192.168.6.195:52050 for unknown channel 3, client has 0 advertised topic(s)
[INFO] [1678937020.104918817] [foxglove_bridge]: Client 192.168.6.195:52050 is advertising "/move_base_simple/goal" (geometry_msgs/PoseStamped) on channel 4
[INFO] [1678937020.105093695] [foxglove_bridge]: Client 192.168.6.195:52050 is no longer advertising /move_base_simple/goal (0 subscribers) on channel 4
[INFO] [1678937020.109083689] [foxglove_bridge]: Client 192.168.6.195:52050 is advertising "/move_base_simple/goal" (geometry_msgs/PoseStamped) on channel 5
[INFO] [1678937020.110877709] [foxglove_bridge]: Client 192.168.6.195:52050 is advertising "/clicked_point" (geometry_msgs/PointStamped) on channel 6
[INFO] [1678937020.111063987] [foxglove_bridge]: Client 192.168.6.195:52050 is no longer advertising /move_base_simple/goal (0 subscribers) on channel 5
[INFO] [1678937020.111169025] [foxglove_bridge]: Client 192.168.6.195:52050 is no longer advertising /clicked_point (0 subscribers) on channel 6

            [rcutils|error_handling.c:108] rcutils_set_error_state()
            This error state is being overwritten:

[...]

PR #233, opened 2023-06-15
Fix QoS history being unknown when copied from existing publisher

achim-k:
### Public-Facing Changes
Fix QoS history being unknown when copied from existing publisher

### Description
When the QoS profile is copied from another existing publisher, it can happen that the history policy is `Unknown`, leading to an error when subsequently trying to create a publisher with that QoS profile. As a fix, we explicitly set the history policy to the system default.

Fixes https://github.com/foxglove/ros-foxglove-bridge/issues/200

### DIFF
--- a/ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
+++ b/ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
@@ -580,11 +580,11 @@ void FoxgloveBridge::clientAdvertise(const foxglove::ClientAdvertisement& advert
                      return endpoint.node_name() != this->get_name() ||
                             endpoint.node_namespace() != this->get_namespace();
                    });
-    const rclcpp::QoS qos = otherPublisherIt == otherPublishers.end()
-                              ? rclcpp::SystemDefaultsQoS()
-                              : rclcpp::QoS(otherPublisherIt->qos_profile())
-                                  // Override the history policy as it can be `unknown`
-                                  .history(rclcpp::HistoryPolicy::SystemDefault);
+    rclcpp::QoS qos = otherPublisherIt == otherPublishers.end() ? rclcpp::SystemDefaultsQoS()
+                                                                : otherPublisherIt->qos_profile();
+    if (qos.history() == rclcpp::HistoryPolicy::Unknown) {
+      qos.history(rclcpp::HistoryPolicy::SystemDefault);
+    }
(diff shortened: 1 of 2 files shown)


ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -582,7 +582,9 @@ void FoxgloveBridge::clientAdvertise(const foxglove::ClientAdvertisement& advert
                    });
     const rclcpp::QoS qos = otherPublisherIt == otherPublishers.end()
                               ? rclcpp::SystemDefaultsQoS()
-                              : otherPublisherIt->qos_profile();
+                              : rclcpp::QoS(otherPublisherIt->qos_profile())
+                                  // Override the history policy as it can be `unknown`
```
defunctzombie, line 586: Would it be better to only override _if_ it is `unknown` and leave it otherwise?

achim-k: Yes, will do. Although it seems that this is ROS distro specific. On `humble` this is always `unknown` (might be a bug), on  `galactic` the history is always set

ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp
```diff
@@ -580,9 +580,11 @@ void FoxgloveBridge::clientAdvertise(const foxglove::ClientAdvertisement& advert
                      return endpoint.node_name() != this->get_name() ||
                             endpoint.node_namespace() != this->get_namespace();
                    });
-    const rclcpp::QoS qos = otherPublisherIt == otherPublishers.end()
-                              ? rclcpp::SystemDefaultsQoS()
-                              : otherPublisherIt->qos_profile();
+    rclcpp::QoS qos = otherPublisherIt == otherPublishers.end() ? rclcpp::SystemDefaultsQoS()
+                                                                : otherPublisherIt->qos_profile();
+    if (qos.history() == rclcpp::HistoryPolicy::Unknown) {
+      qos.history(rclcpp::HistoryPolicy::SystemDefault);
```
jtbandes, line 586: Comment on why this is necessary?

jtbandes (approved):
Is there any possibility to add a test for this?

achim-k:
> Is there any possibility to add a test for this?

Yes, added one. Fails for humble without the history policy fix.

PR #239, opened 2023-06-30
Assume publisher qos depth of 1 if the middleware reports the qos history as unknown

### LINKED ISSUE #238: ROS2 static transforms partially propagating through the web bridge to studio
**Description**

- Version: Foxglove Studio 1.55.0 / Foxglove Web Bridge 0.5.3
- Platform: Foxglove Studio on OSX, Foxglove Web Bridge on Linux
- Data source: Only affects streaming from a live system. Does not appear to affect loading a MCAP file directly into studio.

**Steps To Reproduce**

Take a look at this simple launch file that sets up a basic static transform hierarchy:
https://gist.github.com/asymingt/1d833ac9857ecba36eb2fa523137a7ba

Launch like this from a ROS2 rolling distribution:
```
$ ros2 launch static_frame_issue.launch.py
```

Confirm you see the correct frames with:
```
$ ros2 topic echo /tf_static
transforms:
- header:
    stamp:
      sec: 1687902344
      nanosec: 58697684
    frame_id: bar_child1
  child_frame_id: bar_child1_grandchild
  transform:
    translation:
      x: 0.0
      y: 0.0
      z: 0.0
    rotation:
      x: 0.0
      y: 0.0
      z: 0.0
      w: 1.0
---
transforms:
- header:
    stamp:
      sec: 1687902344
      nanosec: 55601655
    frame_id: bar
  child_frame_id: foo
  transform:
    translation:
      x: 0.0
      y: 0.0
      z: 0.0
    rotation:
      x: 0.0
      y: 0.0
      z: 0.0
      w: 1.0
---
transforms:
- header:
    stamp:
      sec: 1687902344
      nanosec: 54859102
    frame_id: bar
  child_frame_id: bar_child1
  transform:
    translation:
      x: 0.5
      y: 1.0
      z: 2.0
    rotation:
      x: 0.0
      y: 0.0
      z: 0.0
      w: 1.0
---
transforms:
- header:
    stamp:
      sec: 1687902344
      nanosec: 56240568
    frame_id: bar_child2
  child_frame_id: bar_child2_grandchild
  transform:
    translation:
      x: 0.0
      y: 0.0
      z: 0.0
    rotation:
      x: 0.0
      y: 0.0
      z: 0.0
      w: 1.0
---
transforms:
- header:
    stamp:
      sec: 1687902344
      nanosec: 56478622
    frame_id: bar
  child_frame_id: bar_child2
  transform:
    translation:
      x: 0.5
      y: -1.0
      z: 2.0
    rotation:
      x: 0.0
      y: 0.0
      z: 0.0
      w: 1.0
---
```

**Expected Behavior**

I would expect to see a transform hierarchy that looks like this, in both the `Transforms` section of the `3D` panel, as well as in the 3D visualization itself:

```
+ baz
  + bar
    + bar_child1
      + bar_child1_grandchild
    + bar_child2
      + bar_child2_grandchild
    + foo
```

However, only a subset of the frames are actually received and drawn. So, for example sometimes you see this:

```
+ baz
  + bar
    + bar_child1
      + bar_child1_grandchild
```

And sometimes you see this:

```
+ baz
  + bar
    + bar_child2
      + bar_child2_grandchild
```

Example screenshot of the problematic `Transforms` panel:

Screen Shot 2023-06-27 at 2 30 44 PM

Example screenshot of the problematic `3D` window:

Screen Shot 2023-06-27 at 2 32 04 PM

achim-k:
### Public-Facing Changes

Assume publisher qos depth of 1 if the middleware reports the qos history as unknown

### Description
Some RMWs do not report history information of a publisher endpoint in which case the history depth is reported as 0. We internally use this depth to calculate an appropriate depth for the subscription, which uses the `KeepLast(depth)` history policy.
In case a publisher's history depth is reported as 0, we assume a depth of 1 so that the final history depth is at least equal to the number of publishers. This covers the case where there are multiple transient_local publishers with a depth of 1 (e.g. multiple tf_static transform broadcasters). Before this PR, a user would have to manually specify a `min_qos_depth` of N when having N tf_static broadcasters.

This PR also increases the max qos depth default value, 10 seemed very low. I'm not sure anymore why we added an upper limit in the first place. Additionally, a warning is logged when the upper limit is reached.

Fixes #238 
See also https://github.com/foxglove/ros-foxglove-bridge/issues/208

### DIFF
--- a/foxglove_bridge_base/src/test/test_client.cpp
+++ b/foxglove_bridge_base/src/test/test_client.cpp
@@ -12,34 +12,29 @@ namespace foxglove {
 constexpr auto DEFAULT_TIMEOUT = std::chrono::seconds(5);
 
 std::vector<uint8_t> connectClientAndReceiveMsg(const std::string& uri,
-                                                const std::string& topic_name) {
+                                                const std::string& topicName) {
   // Set up text message handler to resolve the promise when the topic is advertised
-  foxglove::Client<websocketpp::config::asio_client> wsClient;
+  auto wsClient = std::make_shared<foxglove::Client<websocketpp::config::asio_client>>();
   std::promise<nlohmann::json> channelPromise;
-  auto channelFuture = channelPromise.get_future();
-  wsClient.setTextMessageHandler([&topic_name, &channelPromise](const std::string& payload) {
-    const auto msg = nlohmann::json::parse(payload);
-    const auto& op = msg.at("op").get<std::string>();
-    if (op == "advertise") {
-      for (const auto& channel : msg.at("channels")) {
-        if (topic_name == channel.at("topic")) {
-          channelPromise.set_value(channel);
-        }
-      }
-    }
-  });
+  auto channelFuture = waitForChannel(wsClient, topicName);
 
   // Connect the client and wait for the channel future
-  if (std::future_status::ready != wsClient.connect(uri).wait_for(DEFAULT_TIMEOUT)) {
+  if (std::future_status::ready != wsClient->connect(uri).wait_for(DEFAULT_TIMEOUT)) {
     throw std::runtime_error("Client failed to connect");
   } else if (std::future_status::ready != channelFuture.wait_for(DEFAULT_TIMEOUT)) {
     throw std::runtime_error("Client failed to receive channel");
   }
 
+  const auto channel = channelFuture.get();
+  const SubscriptionId subscriptionId = 1;
+
   // Set up binary message handler to resolve when a binary message has been received
   std::promise<std::vector<uint8_t>> msgPromise;
   auto msgFuture = msgPromise.get_future();
-  wsClient.setBinaryMessageHandler([&msgPromise](const uint8_t* data, size_t dataLength) {
+  wsClient->setBinaryMessageHandler([&msgPromise](const uint8_t* data, size_t dataLength) {
+    if (ReadUint32LE(data + 1) != subscriptionId) {
+      return;
+    }
     const size_t offset = 1 + 4 + 8;
     std::vector<uint8_t> dataCopy(dataLength - offset);
     std::memcpy(dataCopy.data(), data + offset, dataLength - offset);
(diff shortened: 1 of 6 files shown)


foxglove_bridge_base/src/test/test_client.cpp
```diff
@@ -12,43 +12,37 @@ namespace foxglove {
 constexpr auto DEFAULT_TIMEOUT = std::chrono::seconds(5);
 
 std::vector<uint8_t> connectClientAndReceiveMsg(const std::string& uri,
-                                                const std::string& topic_name) {
+                                                const std::string& topicName) {
   // Set up text message handler to resolve the promise when the topic is advertised
-  foxglove::Client<websocketpp::config::asio_client> wsClient;
+  auto wsClient = std::make_shared<foxglove::Client<websocketpp::config::asio_client>>();
```
jtbandes, line 17: Why does this need to be a shared_ptr? Would unique_ptr work? I don't see the ownership being actually shared anywhere?

foxglove_bridge_base/src/test/test_client.cpp
```diff
@@ -118,6 +112,30 @@ std::future<Service> waitForService(std::shared_ptr<ClientInterface> client,
   return future;
 }
 
+std::future<Channel> waitForChannel(std::shared_ptr<ClientInterface> client,
+                                    const std::string& topicName) {
+  auto promise = std::make_shared<std::promise<Channel>>();
```
jtbandes, line 117: Why shared_ptr? The std::promise example code seems to allocate a promise on the stack and still move it into a thread or lambda.

jtbandes (approved):

foxglove_bridge_base/src/test/test_client.cpp
achim-k: You are right, will see what I can do. Can just pass a raw pointer to `waitForChannel` then

foxglove_bridge_base/src/test/test_client.cpp
achim-k: I remember that John and I had some issues with that but don't remember exactly why. Will see what I can do once I created a new release

PR #246, opened 2023-07-11
Avoid usage of tmpnam() for creating random filename

achim-k:
### Public-Facing Changes

Avoid usage of tmpnam() for creating random filename

### Description
Follow up of #232. ROS build farm complained about usage of `tmpnam`. This PR replaces it by using the epoch time in milliseconds to create a somewhat random filename (used for tests only)

https://build.ros2.org/job/Idev__foxglove_bridge__ubuntu_jammy_amd64/9/

### DIFF
--- a/ros1_foxglove_bridge/tests/smoke_test.cpp
+++ b/ros1_foxglove_bridge/tests/smoke_test.cpp
@@ -1,4 +1,5 @@
 #include <chrono>
+#include <filesystem>
 #include <future>
 #include <thread>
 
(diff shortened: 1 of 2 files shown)


jtbandes (approved):
Why not `mkstemp` as recommended in the lint error?

achim-k:
> Why not `mkstemp` as recommended in the lint error?

Not available under windows. Not a target OS, but it would be good if we could keep foxglove bridge building with robotstack.

PR #247, opened 2023-07-11
Support building with boost asio

achim-k:
### Public-Facing Changes

Support building with boost asio

### Description
Avoids a hard dependency on standalone asio and provides users with more flexibility when using foxglove_bridge in a bigger project. Related: https://github.com/foxglove/ws-protocol/issues/483

Successfully built and tested melodic & noetic with boost asio. I think it would make sense to add a test for this as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,6 @@ project(foxglove_bridge LANGUAGES CXX VERSION 0.6.4)
 
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
-add_definitions(-DASIO_STANDALONE)
 
 macro(enable_strict_compiler_warnings target)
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -179,17 +187,18 @@ if(ROS_BUILD_TYPE STREQUAL "catkin")
       find_package(GTest REQUIRED)
     endif()
     find_package(rostest REQUIRED)
+    find_package(Boost REQUIRED COMPONENTS system)
```
jtbandes, line 190: Should we do this conditionally based on the flags chosen?

jtbandes (approved):
Can we test both of these in CI?

CMakeLists.txt
achim-k: We could, but ROS1 depends on boost anyways, so it does not hurt to have it here

PR #250, opened 2023-07-13
Server code improvements

achim-k:
### Public-Facing Changes

Refactor server code, improve multi-threading safety

### Description
Addresses comments from https://github.com/foxglove/ws-protocol/pull/491. These changes will be applied to that PR once this PR gets merged.

- Fixes modifications to a class member being made while holding only a shared lock
- Log an error if handler functions are not defined by server
- Move text handlers into separate functions
- Other refactoring / simplications

### DIFF
--- a/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
+++ b/foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
@@ -345,7 +345,10 @@ inline void Server<ServerConfiguration>::handleConnectionClosed(ConnHandle hdl) 
     _server.get_alog().write(APP, "Client " + clientName + " unadvertising channel " +
                                     std::to_string(clientChannelId) + " due to disconnect");
     if (_handlers.clientUnadvertiseHandler) {
-      _handlers.clientUnadvertiseHandler(clientChannelId, hdl);
+      try {
+        _handlers.clientUnadvertiseHandler(clientChannelId, hdl);
+      } catch (...) {
+      }
     }
   }
 
@@ -358,7 +361,10 @@ inline void Server<ServerConfiguration>::handleConnectionClosed(ConnHandle hdl) 
   if (_handlers.unsubscribeHandler) {
     for (const auto& [chanId, subs] : oldSubscriptionsByChannel) {
       (void)subs;
-      _handlers.unsubscribeHandler(chanId, hdl);
+      try {
+        _handlers.unsubscribeHandler(chanId, hdl);
+      } catch (...) {
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -38,6 +38,29 @@
     }                                                                                          \
   }
 
+namespace {
+
+constexpr uint32_t Integer(const std::string_view str) {
```
jtbandes, line 43: this function should probably be named StringHash or something that actually explains what it does.

Is there any compile time guarantee that the strings we use it with actually hash to different values? I guess we would hope to get a compile error if it resulted in a switch with multiple identical `case`s?

edit: but I see we are not using switch..

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -345,7 +372,10 @@ inline void Server<ServerConfiguration>::handleConnectionClosed(ConnHandle hdl)
     _server.get_alog().write(APP, "Client " + clientName + " unadvertising channel " +
                                     std::to_string(clientChannelId) + " due to disconnect");
     if (_handlers.clientUnadvertiseHandler) {
-      _handlers.clientUnadvertiseHandler(clientChannelId, hdl);
+      try {
+        _handlers.clientUnadvertiseHandler(clientChannelId, hdl);
+      } catch (...) {
+      }
```
jtbandes, line 378: Why do nothing here? Seems like we should log an error if possible?

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -605,332 +621,46 @@ inline void Server<ServerConfiguration>::handleTextMessage(ConnHandle hdl, Messa
...
-      } catch (const std::exception& e) {
-        sendStatusAndLogMsg(hdl, StatusLevel::Error, e.what());
-      } catch (...) {
-        sendStatusAndLogMsg(hdl, StatusLevel::Error, op + ": Failed to execute handler");
-      }
-    } break;
-    default: {
+  try {
+    const auto intOp = Integer(op);
+    if (intOp == SUBSCRIBE) {
```
jtbandes, line 633: can this be a switch?

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
```diff
@@ -1391,6 +1121,312 @@ inline bool Server<ServerConfiguration>::hasCapability(const std::string& capabi
          _options.capabilities.end();
 }
 
+template <typename ServerConfiguration>
+inline bool Server<ServerConfiguration>::hasHandler(uint32_t op) const {
+  switch (op) {
+    case SUBSCRIBE:
+      return static_cast<bool>(_handlers.subscribeHandler);
```
jtbandes, line 1128: nit: you can just do `bool(...)`

jtbandes (approved):
👍

foxglove_bridge_base/include/foxglove_bridge/websocket_server.hpp
achim-k: Renamed it and changed it back to a switch

PR #251, opened 2023-07-14
Fix segfault due to invalid iterator

achim-k:
### Public-Facing Changes

Fix segfault due to invalid iterator

### Description
Fixes a segfault caused by using an invalid iterator as argument to `std::unordered_map<...>::erase`. This regression was introduced in #250.


jtbandes (approved):
Can we add a test for this easily?

achim-k:
> Can we add a test for this easily?

Yes! Will do, noticed that we currently do not test the `unsubscribe` operation.

PR #270, opened 2023-10-24
Fix exponential regex backtracking issue with default `asset_uri_whitelist`

achim-k:
### Public-Facing Changes

Fix exponential regex backtracking issue with default `asset_uri_whitelist`

### Description
When fetching the asset `package://foxglove_simulation/meshes/kinect_jpg` I noticed that I never got a response. It turned out that the reason for this was due to catastrophic backtracking caused by the default `asset_uri_whitelist` regular expression pattern. The default pattern contains nested quantifiers with alternations, which can lead to exponential backtracking. This can cause the program to hang or freeze when trying to match certain input strings.

This PR changes the default `asset_uri_whitelist` regex pattern to avoid catastrophic backtracking. It additionally adds `png`, `jpg` and `jpeg` to the allowed extensions to allow for retrieval of texture files.


jtbandes (approved):
👏

jhurliman:
Add TIFF and WEBP as well?

README.md
```diff
@@ -77,7 +77,7 @@ Parameters are provided to configure the behavior of the bridge. These parameter
...
+ * __asset_uri_allowlist__: List of regular expressions ([ECMAScript grammar](https://en.cppreference.com/w/cpp/regex/ecmascript)) of allowed asset URIs. Uses the [resource_retriever](https://index.ros.org/p/resource_retriever/github-ros-resource_retriever) to resolve `package://`, `file://` or `http(s)://` URIs. Note that this list should be carefully configured such that no confidential files are accidentally exposed over the websocket connection. As an extra security measure, URIs containing two consecutive dots (`..`) are disallowed as they could be used to construct URIs that would allow retrieval of confidential files if the allowlist is not configured strict enough (e.g. `package://<pkg_name>/../../../secret.txt`). Defaults to `["^package://(?:\w+/)*\w+\.(?:dae|stl|urdf|xacro|png|jpg|jpeg)$"]`.
```
jhurliman, line 80: Would it make sense to add obj, mtl, fbx, glb, gltf to this default list?

achim-k: Yeah, we could do that. Plus tiff and webp?

jhurliman: Yes please!

jtbandes: I thought Chrome didn't support tiff: https://stackoverflow.com/a/2177012/23649

jtbandes: Do these mtl, fbx, etc. files all have separate three.js loaders? I doubt they would be supported in studio out of the box, right?

jtbandes (approved):
Might want to add tif as an alias for tiff.

README.md
jhurliman: > I thought Chrome didn't support tiff: https://stackoverflow.com/a/2177012/23649

There is a special loader in three.js for TIFF support

jhurliman: > Do these mtl, fbx, etc. files all have separate three.js loaders? I doubt they would be supported in studio out of the box, right?

Yes. MTL files accompany OBJ, and are supported by three.js along with FBX and GLTF (text and binary)

PR #274, opened 2023-11-14
Fix delayed notification of clients about client published topic [ROS 1]

### LINKED ISSUE #273: Send message schema immediately after Client Advertise
**Description**

I'm trying to replace my usage of rosbridge with foxglove-bridge because of its faster performance. I publish topics from my client application, which typically have no other publishers. I'd like to receive the schema of my published type immediately after advertising so that I may publish messages immediately. I would prefer not to include the schema in my client code so that custom messages can be updated freely and unfilled fields will be zero-initialized. rosbridge works around this problem because it serializes in the server, which has access to all message definitions. I would prefer to avoid a solution where I `fetchAsset` for the message definitions.

- Version: 0.7.2-1focal.20231013.185151
- Platform: Ubuntu 20.04, ROS Noetic

**Steps To Reproduce**
```js

client.on('advertise', (channels) => {
	for (const channel of channels) {
		console.log(channel.schemaName, channel.schema.split('\n')[0] + '…');
	}
});

client.on('open', () => {
	client.advertise({
		topic: '/topic',
		encoding: 'ros1',
		schemaName: 'geometry_msgs/Polygon',
	});
});
```

Notice how it takes a few seconds for the `geometry_msgs/Polygon` schema to be logged—up to `max_update_ms`.

**Expected Behavior**

The `client.on('advertise')` callback should be called immediately (or soon after—certainly not `max_update_ms` long) after advertising a topic, or at least if a definition for the `schemaName` has never been sent to the client.

achim-k:
### Public-Facing Changes

Fix delayed notification of clients about client published topic [ROS 1]

### Description
We manually trigger topic discovery to avoid the delay that would incur when waiting for the next timer-triggered topic discovery. Only necessary for the ROS 1 implementation as the ROS 2 node listens for graph changes.

Fixes #273
Resolves FG-5729

### DIFF
--- a/ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
+++ b/ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
@@ -345,6 +345,8 @@ private:
       ROS_INFO("Client %s is advertising \"%s\" (%s) on channel %d",
                _server->remoteEndpointString(clientHandle).c_str(), channel.topic.c_str(),
                channel.schemaName.c_str(), channel.channelId);
+      // Trigger topic discovery so other clients are immediately informed about this new topic.
+      updateAdvertisedTopics();
     } else {
       const auto errMsg =
         "Failed to create publisher for topic " + channel.topic + "(" + channel.schemaName + ")";


jtbandes (approved):
Is there an equivalent change we should make for ROS 2?

achim-k:
> Is there an equivalent change we should make for ROS 2?

Not necessary: 
> Only necessary for the ROS 1 implementation as the ROS 2 node listens for graph changes.

PR #278, opened 2023-12-12
Solved bug with incompatible QoS policies (second try)

tudoroancea:
### Public-Facing Changes

Solved bug with incompatible QoS policies

### Description

Fixes the following [issue](https://github.com/tudoroancea/ros-foxglove-bridge/issues/1)

### DIFF
--- a/ros2_foxglove_bridge/src/parameter_interface.cpp
+++ b/ros2_foxglove_bridge/src/parameter_interface.cpp
@@ -94,7 +94,7 @@ static foxglove::Parameter fromRosParam(const rclcpp::Parameter& p) {
   } else if (type == rclcpp::ParameterType::PARAMETER_BOOL_ARRAY) {
     std::vector<foxglove::ParameterValue> paramVec;
     for (const auto value : p.as_bool_array()) {
-      paramVec.push_back(value);
+      paramVec.push_back(foxglove::ParameterValue(value));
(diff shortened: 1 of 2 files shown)


ros2_foxglove_bridge/src/parameter_interface.cpp
```diff
@@ -94,7 +94,7 @@ static foxglove::Parameter fromRosParam(const rclcpp::Parameter& p) {
   } else if (type == rclcpp::ParameterType::PARAMETER_BOOL_ARRAY) {
     std::vector<foxglove::ParameterValue> paramVec;
     for (const auto value : p.as_bool_array()) {
-      paramVec.push_back(value);
+      paramVec.push_back(foxglove::ParameterValue(value));
```
achim-k, line 97: This change was already commited in https://github.com/foxglove/ros-foxglove-bridge/pull/277, can you remove it from this PR? Not sure why it isn't shown as conflicting.

achim-k (approved):

ros2_foxglove_bridge/src/parameter_interface.cpp
tudoroancea: I based this branch on the one I created from #277 (to be able to compile the code and test it on my mac), that's why the same commit appears

PR #285, opened 2024-02-12
Remove dual ROS 1+2 devcontainer, remove ROS Galactic from the support matrix

jhurliman:
### Public-Facing Changes

- ROS Galactic support has been dropped from the support matrix

### Description

In anticipation of adding ROS 2-specific functionality that relies on ROS Humble or later, ROS Galactic has been removed from the support matrix. The dual ROS 1+2 devcontainer has been replaced with a ROS 2 Humble devcontainer that matches what is used in CI.


.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ros_distribution: [melodic, noetic, galactic, humble, iron, rolling]
+        ros_distribution: [melodic, noetic, humble, iron, rolling]
```
achim-k, line 24: maybe we should leave that? Technically people could still manually build for `galactic`

.devcontainer/devcontainer.json
```diff
@@ -32,7 +31,5 @@
   },
   "capAdd": [
     "SYS_PTRACE"
-  ],
-  // Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
-  "remoteUser": "vscode"
```
achim-k, line 37: What's the impact of removing this? I think files will be owned by root which messes with permissions on the host system, not sure anymore.

achim-k (commented):
> In anticipation of adding ROS 2-specific functionality that relies on ROS Humble or later, ROS Galactic has been removed from the support matrix.

Does that mean that your anticipated changes will prevent users from manually building foxglove_bridge for galactic?

achim-k (approved):

jhurliman:
> > In anticipation of adding ROS 2-specific functionality that relies on ROS Humble or later, ROS Galactic has been removed from the support matrix.
> 
> Does that mean that your anticipated changes will prevent users from manually building foxglove_bridge for galactic?

Yes, that's correct. Once `ros2_babel_fish` code is vendored into this repo, anyone trying to build `ros-foxglove-bridge` against a ROS 2 release before Humble will see compilation errors about trying to call methods that don't exist.

PR #288, opened 2024-02-23
[WIP] ROS 2 JSON publishing support

jhurliman:
### Public-Facing Changes
- Clients to the ROS 2 bridge can publish messages as JSON instead of `cdr`. JSON payloads will be re-serialized to `cdr` by the bridge before publishing to the ROS 2 graph

### Description
This PR vendors in the unpublished <https://github.com/LOEWE-emergenCITY/ros2_babel_fish> library and uses it to support JSON message publishing from WebSocket clients. Incoming JSON messages are parsed, transformed into native ROS 2 messages field-by-field, then published to the ROS 2 graph.

### TODO
- [x] Test with empty `{}` JSON message, simple message types
- [ ] Test compound messages, `Time`/`Duration` fields, etc
- [x] Array support
- [ ] Solve ROS Iron / Rolling support (disable with #ifdef?)
- [x] Add test coverage
- [ ] Update docs


ros2_foxglove_bridge/src/json_to_ros.cpp
```diff
@@ -0,0 +1,139 @@
...
+      auto& compound = field.as<ros2_babel_fish::CompoundMessage>();
+      for (const auto& key : compound.keys()) {
+        assignJsonToMessageField(compound[key], value[key]);
+      }
+      break;
+    }
+    case MessageType::Array: {
+      // Ensure the JSON value is an array
+      if (!value.is_array()) {
+        break;
```
achim-k, line 79: shouldn't we throw an exception in that case?

ros2_foxglove_bridge/src/json_to_ros.cpp
```diff
@@ -0,0 +1,139 @@
...
+    if (!json.contains(key)) {
+      continue;
+    }
+
+    try {
+      assignJsonToMessageField(rosMsg[key], json[key]);
+    } catch (const std::exception& e) {
+      return e;
+    }
+  }
```
achim-k, line 131: that's a lot of try-catch blocks. Having only one try-catch block should make it more readable

achim-k: Or related to my previous comment: Don't catch exceptions here and return the output message rather than the std::optional

ros2_foxglove_bridge/tests/json_to_ros_test.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include <foxglove_bridge/json_to_ros.hpp>
+
+TEST(JsonToRosTest, EmptyStringMsg) {
+  const std::string payload = "{}";
+
+  auto babelFish = ros2_babel_fish::BabelFish::make_shared();
+
+  ros2_babel_fish::CompoundMessage::SharedPtr output;
+  auto res = foxglove_bridge::jsonMessageToRos(payload, "std_msgs/msg/String", babelFish, output);
```
achim-k, line 11: Can you add a test for a more complex message type? E.g. `geometry_msgs/PoseStamped` or something similar, ideally containing an array.

achim-k (commented):
It would be nice if we could avoid having to copy all the babel fish code. The `humble` release is currently blocked tho (see https://github.com/LOEWE-emergenCITY/ros2_babel_fish/issues/7). Maybe we could also avoid copying all that code with some git submodule / cmake magic?

In general, this approach (with babel fish) should also work for more recent distros (iron, rolling), right?

jhurliman:
> It would be nice if we could avoid having to copy all the babel fish code. The `humble` release is currently blocked tho (see [LOEWE-emergenCITY/ros2_babel_fish#7](https://github.com/LOEWE-emergenCITY/ros2_babel_fish/issues/7)).

"Currently" is underselling it. Now that this module has been superseded by native ROS 2 API methods in Iron+, it is deprecated and will never be released.

> Maybe we could also avoid copying all that code with some git submodule / cmake magic?

My understanding is git submodules and cmake magic are generally frowned upon for ROS / ROS 2 projects that will be built and distributed in the ROS build system and third party developers. If you have a reference example in mind I can take a look.

> In general, this approach (with babel fish) should also work for more recent distros (iron, rolling), right?

No, babel fish does not build in Iron+ and has been replaced by API methods. It will require a different (albeit better) approach.

achim-k:
> > Maybe we could also avoid copying all that code with some git submodule / cmake magic?
> 
> My understanding is git submodules and cmake magic are generally frowned upon for ROS / ROS 2 projects that will be built and distributed in the ROS build system and third party developers. If you have a reference example in mind I can take a look.

The `mcap_vendor` package uses [`FetchContent`](https://cmake.org/cmake/help/latest/module/FetchContent.html):
https://github.com/ros2/rosbag2/blob/bee10b49e6946c9467a421898100f5c93a49b246/mcap_vendor/CMakeLists.txt#L35-L47

StefanFabian:
> > In general, this approach (with babel fish) should also work for more recent distros (iron, rolling), right?
> 
> No, babel fish does not build in Iron+ and has been replaced by API methods. It will require a different (albeit better) approach.

Just read this by chance and would like to say that it does in fact build in jazzy after some small changes addressing API changes. I'm also interested in the API methods that replace its functionality as I couldn't find anything about that.
There's a `GenericSubscription` and `GenericPublisher` now but they seem to only work with raw data messages which is the smallest issue `ros2_babel_fish` had to solve.
You still have to load the required message library and use the introspection api (which has been around since foxy or earlier) to interpret these messages which is the main challenge `ros2_babel_fish` is simplifying.
If there's an easier API now, I would be very interested in that.

ros2_foxglove_bridge/include/foxglove_bridge/ros2_foxglove_bridge.hpp
```diff
@@ -20,6 +20,10 @@
 #include <foxglove_bridge/server_factory.hpp>
 #include <foxglove_bridge/utils.hpp>
 
+#ifdef ENABLE_JSON_MESSAGES
+#include <ros2_babel_fish/babel_fish.hpp>
+#endif
```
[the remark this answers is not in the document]
facontidavide: I agree, forward declaration might be desirable. Altenrative, we can use the **Pimpl** idiom.

Just one additional question: in which cases will a user prefer to compile this with ENABLE_JSON_MESSAGES undefined?

facontidavide:
What is blocking this PR from being merged? @StefanFabian correct me if I am wrong, but ros2_bable_fish haven't been released as a package, right? Maybe we should start fixing that?

StefanFabian:
No, it hasn't been released yet.
As jazzy is the first distro (except for foxy which I find kind of funny) where my fixes are merged (or LTS distro, not sure if it's already in Iron), I didn't release it for the older distros.
We are starting the switch to ROS2 in our team this week and I will prepare the release in the next weeks.
I can also release it for the older distros when I do so, you don't need the part affected by the PRs anyway.

defunctzombie:
Replaced by https://github.com/foxglove/ros-foxglove-bridge/pull/307

PR #293, opened 2024-04-23
Fix service definition parsing on ROS rolling

achim-k:
### Changelog
Fix service definition parsing on ROS rolling

### Description
Prior to this change, we used the auto-generated `_Request.msg` and `_Response.msg` message definitions for parsing a service's request and response definition. However, these message definitions have been removed in https://github.com/ros2/rosidl/pull/753 causing foxglove bridge to raise warnings that the service definition could not be found. This PR changes the service definition parsing to lookup the service definition using the `.srv` file and splitting the content using the `===` seperator to obtain the request and response definition.

### DIFF
--- a/ros2_foxglove_bridge/include/foxglove_bridge/utils.hpp
+++ b/ros2_foxglove_bridge/include/foxglove_bridge/utils.hpp
@@ -30,7 +30,7 @@ inline std::vector<std::string> splitMessageDefinitions(std::istream& stream) {
   std::string line = "";
   std::string definition = "";
 
-  while (getline(stream, line)) {
+  while (std::getline(stream, line)) {
     if (line == "---") {
       definitions.push_back(trimString(definition));
       definition = "";
--- a/ros2_foxglove_bridge/include/foxglove_bridge/utils.hpp
+++ b/ros2_foxglove_bridge/include/foxglove_bridge/utils.hpp
@@ -17,4 +17,30 @@ inline std::pair<std::string, std::string> getNodeAndNodeNamespace(const std::st
   return std::make_pair(fqnNodeName.substr(0, found), fqnNodeName.substr(found + 1));
 }
 
+inline std::string trimString(std::string& str) {
+  constexpr char whitespaces[] = "\t\n\r ";
+  str.erase(0, str.find_first_not_of(whitespaces));  // trim left
+  str.erase(str.find_last_not_of(whitespaces) + 1);  // trim right
+  return str;
+}
+
+inline std::vector<std::string> splitMessageDefinitions(std::istream& stream) {
+  std::vector<std::string> definitions;
+
+  std::string line = "";
+  std::string definition = "";
+
+  while (getline(stream, line)) {
+    if (line == "---") {
+      definitions.push_back(trimString(definition));
+      definition = "";
+    } else {
+      definition += line + "\n";
+    }
+  }
+
+  definitions.push_back(trimString(definition));
+  return definitions;
+}
+
 }  // namespace foxglove_bridge
--- a/ros2_foxglove_bridge/src/message_definition_cache.cpp
+++ b/ros2_foxglove_bridge/src/message_definition_cache.cpp
@@ -31,6 +31,8 @@ static const std::unordered_set<std::string> PRIMITIVE_TYPES{
   "bool",  "byte",   "char",  "float32", "float64", "int8",   "uint8",
   "int16", "uint16", "int32", "uint32",  "int64",   "uint64", "string"};
 
+constexpr char SEP[] = "---\n";
+
 static std::set<std::string> parse_msg_dependencies(const std::string& text,
                                                     const std::string& package_context) {
   std::set<std::string> dependencies;
@@ -130,8 +132,6 @@ static std::string trim_string(std::string& str) {
 /// @return A tuple holding goal, result and feedback definitions
 static std::tuple<std::string, std::string, std::string> split_action_msg_definition(
   const std::string& action_definition) {
-  constexpr char SEP[] = "---";
-
   auto definitions = split_string(action_definition, SEP);
   if (definitions.size() != 3) {
     throw std::invalid_argument("Invalid action definition:\n" + action_definition);
@@ -145,8 +145,6 @@ static std::tuple<std::string, std::string, std::string> split_action_msg_defini
 /// @return A tuple holding request and response definitions
 static std::tuple<std::string, std::string> split_service_msg_definition(
   const std::string& service_definition) {
-  constexpr char SEP[] = "---";
-
   auto definitions = split_string(service_definition, SEP);
   if (definitions.size() != 2) {
     throw std::invalid_argument("Invalid service definition:\n" + service_definition);
--- a/ros2_foxglove_bridge/src/message_definition_cache.cpp
+++ b/ros2_foxglove_bridge/src/message_definition_cache.cpp
@@ -226,14 +226,16 @@ const MessageSpec& MessageDefinitionCache::load_message_spec(
   }
 
   if (subfolder == "action") {
-    const auto splitDefinitions = foxglove_bridge::splitMessageDefinitions(file);
-    if (splitDefinitions.size() != 3) {
-      throw std::invalid_argument("Invalid action definition: " + filename);
+    const auto split_definitions = foxglove_bridge::splitMessageDefinitions(file);
+    if (split_definitions.size() != 3) {
+      throw std::invalid_argument("Invalid action definition in " + filename +
+                                  ": Expected 3 definitions, got " +
+                                  std::to_string(split_definitions.size()));
     }
 
-    const auto& goalDef = splitDefinitions[0];
-    const auto& resultDef = splitDefinitions[1];
-    const auto& feedbackDef = splitDefinitions[2];
+    const auto& goalDef = split_definitions[0];
+    const auto& resultDef = split_definitions[1];
+    const auto& feedbackDef = split_definitions[2];
 
     // Define type definitions for each action subtype.
     // These type definitions may include additional fields such as the goal_id.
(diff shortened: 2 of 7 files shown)


ros2_foxglove_bridge/src/message_definition_cache.cpp
```diff
@@ -133,6 +133,21 @@ static std::tuple<std::string, std::string, std::string> split_action_msg_defini
...
 
+/// @brief Split an service definition into individual request and response definitions.
+/// @param service_definition The full services definition as read from a .srv file
+/// @return A tuple holding request and response definitions
+static std::tuple<std::string, std::string> split_service_msg_definition(
+  const std::string& service_definition) {
+  constexpr char SEP[] = "---\n";
+
+  const auto definitions = split_string(service_definition, SEP);
+  if (definitions.size() != 2) {
```
jtbandes, line 144: Is it legal for the response to be totally empty (no newline) e.g. something like `"int a\n---\nint b\n---"` ?

ros2_foxglove_bridge/src/message_definition_cache.cpp
```diff
@@ -213,44 +254,61 @@ const MessageSpec& MessageDefinitionCache::load_message_spec(
...
-      // Create a MessageSpec instance for every action subtype and add it to the cache.
-      for (const auto& [action_suffix, definition] : action_type_definitions) {
-        DefinitionIdentifier definition_id;
-        definition_id.format = definition_identifier.format;
-        definition_id.package_resource_name = package + "/action/" + action_name + action_suffix;
-        msg_specs_by_definition_identifier_.emplace(
-          definition_id, MessageSpec(definition_id.format, definition, package));
-      }
+  if (subfolder == "action") {
+    const auto [goalDef, resultDef, feedbackDef] = split_action_msg_definition(contents);
```
jtbandes, line 258: Interesting – didn't realize we were already doing a manual split for .action but not .srv. Why were they different?

achim-k: For services it was much easier to use the `_Request` and `_Response` msg definitions which basically didn't need any additional code. For action types, these auto-generated msg definitions didn't exist.

ros2_foxglove_bridge/src/message_definition_cache.cpp
achim-k: Yes, I believe that's legal, also for the request. I will test if that's an issue

ros2_foxglove_bridge/src/message_definition_cache.cpp
```diff
@@ -118,19 +118,41 @@ static std::vector<std::string> split_string(const std::string& str,
...
 /// @brief Split an action definition into individual goal, result and feedback definitions.
 /// @param action_definition The full action definition as read from a .action file
 /// @return A tuple holding goal, result and feedback definitions
 static std::tuple<std::string, std::string, std::string> split_action_msg_definition(
   const std::string& action_definition) {
-  constexpr char SEP[] = "\n---\n";
+  constexpr char SEP[] = "---";
 
-  const auto definitions = split_string(action_definition, SEP);
+  auto definitions = split_string(action_definition, SEP);
```
jtbandes, line 135: Now is there any risk that this conflict with a default value, e.g. will `string STR="foo---bar"` break it?

ros2_foxglove_bridge/src/message_definition_cache.cpp
achim-k: "---\n" is correct:

> A service file contains two message definitions which are separated by a line which only contains three dashes:

achim-k: However, colcon doesn't raise an error if the service definition ends with `---` without final LF. I will have to cover that case as well :sob:

ros2_foxglove_bridge/include/foxglove_bridge/utils.hpp
```diff
@@ -17,4 +17,30 @@ inline std::pair<std::string, std::string> getNodeAndNodeNamespace(const std::st
...
+  return str;
+}
+
+inline std::vector<std::string> splitMessageDefinitions(std::istream& stream) {
+  std::vector<std::string> definitions;
+
+  std::string line = "";
+  std::string definition = "";
+
+  while (getline(stream, line)) {
```
jtbandes, line 33:
```suggestion
  while (std::getline(stream, line)) {
```

ros2_foxglove_bridge/src/message_definition_cache.cpp
```diff
@@ -211,59 +224,89 @@ const MessageSpec& MessageDefinitionCache::load_message_spec(
...
-        throw DefinitionNotFoundError(definition_identifier.package_resource_name);
-      }
-      return it->second;
-    } else {
-      RCUTILS_LOG_ERROR_NAMED("foxglove_bridge",
-                              "Action IDL definitions are currently not supported");
+  if (subfolder == "action") {
+    const auto splitDefinitions = foxglove_bridge::splitMessageDefinitions(file);
+    if (splitDefinitions.size() != 3) {
+      throw std::invalid_argument("Invalid action definition: " + filename);
```
jtbandes, line 231: nit: would be helpful to include the expected & actual number of definitions for debugging

ros2_foxglove_bridge/src/message_definition_cache.cpp
```diff
@@ -211,59 +224,89 @@ const MessageSpec& MessageDefinitionCache::load_message_spec(
...
+    // Find the the subtype that was originally requested and return it.
+    const auto it = msg_specs_by_definition_identifier_.find(definition_identifier);
+    if (it == msg_specs_by_definition_identifier_.end()) {
+      throw DefinitionNotFoundError(definition_identifier.package_resource_name);
+    }
+    return it->second;
+  } else if (subfolder == "srv") {
+    const auto splitDefinitions = foxglove_bridge::splitMessageDefinitions(file);
+    if (splitDefinitions.size() != 2) {
+      throw std::invalid_argument("Invalid service definition: " + filename);
```
jtbandes, line 271: nit: would be helpful to include the expected & actual number of definitions for debugging

ros2_foxglove_bridge/src/message_definition_cache.cpp
```diff
@@ -177,14 +185,19 @@ const MessageSpec& MessageDefinitionCache::load_message_spec(
...
-                                 ? action_name + ".action"
-                                 : type_name + extension_for_format(definition_identifier.format);
+  std::string filename = "";
+  if (subfolder == "action") {
+    // The action type name includes the subtype which we have to remove to get the action name.
+    // Type name: Fibonacci_FeedbackMessage -> Action name: Fibonacci
+    filename = remove_action_subtype(type_name) + ".action";
+  } else if (subfolder == "srv") {
+    // The service type name includes the subtype which we have to remove to get the service name.
+    // Type name: SetBool_Request -> Service name: SetBool
```
jtbandes, line 196: I thought these don't exist anymore? Does the code have to handle both with & without the suffix because they exist in <=iron but not in >iron?

ros2_foxglove_bridge/src/message_definition_cache.cpp
```diff
@@ -211,59 +224,89 @@ const MessageSpec& MessageDefinitionCache::load_message_spec(
...
+    }
+
+    // Find the the subtype that was originally requested and return it.
+    const auto it = msg_specs_by_definition_identifier_.find(definition_identifier);
+    if (it == msg_specs_by_definition_identifier_.end()) {
+      throw DefinitionNotFoundError(definition_identifier.package_resource_name);
+    }
+    return it->second;
+  } else if (subfolder == "srv") {
+    const auto splitDefinitions = foxglove_bridge::splitMessageDefinitions(file);
```
jtbandes, line 269: nit: inconsistent snake vs camel case

ros2_foxglove_bridge/tests/smoke_test.cpp
```diff
@@ -537,6 +537,24 @@ TEST_F(ParameterTest, testGetParametersParallel) {
...
+
+  EXPECT_EQ(service.name, SERVICE_NAME);
+  EXPECT_EQ(service.type, "std_srvs/srv/SetBool");
+  EXPECT_EQ(service.requestSchema, "bool data # e.g. for hardware enabling / disabling");
+  EXPECT_EQ(service.responseSchema,
+            "bool success   # indicate successful run of triggered service\nstring message # "
+            "informational, e.g. for error messages");
+
+  std_srvs::srv::SetBool::Request requestMsg;
+  requestMsg.data = true;
```
jtbandes, line 555: what's the point of these lines?

ros2_foxglove_bridge/tests/utils_test.cpp
jtbandes, line 1: 👍

jtbandes (approved):

ros2_foxglove_bridge/src/message_definition_cache.cpp
jhurliman: This follows Google C++ style convention, using snake case for namespaces and camel case for method names

jtbandes: I meant the local variable name, it seems we have local variables in both styles

ros2_foxglove_bridge/src/message_definition_cache.cpp
achim-k: We still use `_Request` and `_Response` as type name suffixes when advertising the services as we can't (or rather shouldn't) advertise request and response types with the same type name.

ros2_foxglove_bridge/tests/smoke_test.cpp
achim-k: copy paste error it seems

Issue #301, opened 2024-06-11
Invalid service definition


tonynajjar:
**Description**

- Version of foxglove bridge: 0.7.7
- Platform: foxglove studio 2.6.0

**Steps To Reproduce**

```
[foxglove_bridge-22] [WARN] [1718109515.363661001] [foxglove_bridge]: Failed to add service /perception/front_left/device_info;(realsense2_camera_msgs/srv/DeviceInfo): Invalid service definition in DeviceInfo.srv: Expected 2 definitions, got 1
```

The srv seems reasonable https://github.com/IntelRealSense/realsense-ros/blob/ros2-development/realsense2_camera_msgs/srv/DeviceInfo.srv

achim-k:
Yeah, that's strange. I will take a look, thanks for reporting

tonynajjar:
I'm on Humble. I'm reverting your commit and checking if it reproduces

tonynajjar:
confirmed, checked out the commit before the PR I linked (hash 3e35af32fde80dc42e6c7bea40a2cc32c30170f4) and the issue does not reproduce

achim-k:
The reason is an additional CR `0x0D` before the actual line feed `0x0A` at the end of the line :sob: Will push a fix in a couple of hours.

![image](https://github.com/foxglove/ros-foxglove-bridge/assets/9250155/346eaeaa-1923-4592-9b8e-b06d57fa3585)

tonynajjar:
It's impressive how fast you figured out such a hidden issue :laughing: Let me know when you have the fix, I might be still in time to give it a quick test

PR #303, opened 2024-06-11
Fix srv definition parsing failing due to carriage return

achim-k:
### Changelog
Fix srv definition parsing failing due to carriage return

### Docs
None

### Description
Fixes srv (or action) definition parsing failing when the definition seperator `---` is terminated with `\r\n` instead of just `\n`. More details in #301.

Fixes: #301


achim-k:
cc @tonynajjar

achim-k:
Rolling CI tests are still failing -> #304

jtbandes (approved):
Do we support \r\n in other places or only after the `---` separator?

achim-k:
This was the only place where we did a string comparison, for other places we do not have to handle different line endings.

PR #309, opened 2024-06-26
Fix rolling smoke tests crashing

### LINKED ISSUE #304: Rolling smoke test is crashing on exit
On rolling, the smoke test crashes during destruction of the executor, causing a test failure. Something my have changed upstream that causes this behavior. We have to fix this such that PR status checks are working reliably again.

achim-k:
### Changelog
Fix rolling test crashes

### Docs
None

### Description
Fixes the rolling smoke tests crashing due to incorrect executor cleanup. In the end, the only thing missing was a `executor.reset();` at the end of `main`. However I took the opportunity to clean up a little.

Fixes #304

### DIFF
--- a/ros2_foxglove_bridge/tests/smoke_test.cpp
+++ b/ros2_foxglove_bridge/tests/smoke_test.cpp
@@ -4,11 +4,11 @@
 #include <thread>
 
 #include <gtest/gtest.h>
-#include <rclcpp_components/component_manager.hpp>
 #include <std_msgs/msg/string.hpp>
 #include <std_srvs/srv/set_bool.hpp>
 #include <websocketpp/config/asio_client.hpp>
 
+#include <foxglove_bridge/ros2_foxglove_bridge.hpp>
 #include <foxglove_bridge/test/test_client.hpp>
 #include <foxglove_bridge/websocket_client.hpp>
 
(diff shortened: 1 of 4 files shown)


ros2_foxglove_bridge/tests/smoke_test.cpp
```diff
@@ -759,6 +757,7 @@ int main(int argc, char** argv) {
   const auto testResult = RUN_ALL_TESTS();
   executor->cancel();
   spinnerThread.join();
+  executor.reset();
```
jtbandes, line 760: why is this necessary? this is just a shared_ptr right? so even though we already called cancel(), it is problematic to have the object's lifetime last beyond the shutdown() call? Probably could use a comment here because this line seems fairly useless otherwise.

If the lifetime matters so much, can we use `make_unique` instead of `make_shared`?

jtbandes (approved):

ros2_foxglove_bridge/tests/smoke_test.cpp
achim-k: I think the component manager was still holding on to it which caused issues. I have simplified this now quite a lot and am using the node directly instead of loading it via the component manager.

ros2_foxglove_bridge/tests/smoke_test.cpp
```diff
@@ -732,9 +732,7 @@ int main(int argc, char** argv) {
   testing::InitGoogleTest(&argc, argv);
   rclcpp::init(argc, argv);
 
-  const size_t numThreads = 2;
-  auto executor =
-    rclcpp::executors::MultiThreadedExecutor::make_shared(rclcpp::ExecutorOptions{}, numThreads);
+  auto executor = rclcpp::executors::SingleThreadedExecutor::make_shared();
```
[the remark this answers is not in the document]
achim-k: no good reason actually, just noticed that it also works with a single threaded executor so I left it

PR #316, opened 2024-07-11
Make ROS1 service type retrieval more robust

### LINKED ISSUE #306: Assertion error on read_callback_ in roscpp for ROS1 bridge
**Description**

We're getting an assertion error when running the ROS1 bridge

```log
FATAL] [1716982265.933923467]: ASSERTION FAILED
        file = ./clients/roscpp/src/libros/connection.cpp
        line = 274
        cond = !read_callback_
```

* Version: 0.77 (15df09892cc2cbea0dd6efc8c53e4f01e0881511)
* Platform: linux x86_64 (ubuntu jammy)

**Steps To Reproduce**

Compile and start the ROS1 bridge.

**Expected Behavior**

This assertion should not be triggered.

We found that when debugging the assertion is not immediately triggered. If we add "sufficient" sleep before the call to `getConnection()`, the assertion is not triggered, and the ROS1 bridge works, we can get a connection. However at some point later in time we still get the same assertion.

```diff
--- ros1_foxglove_bridge/src/service_utils.cpp
+++ ros1_foxglove_bridge/src/service_utils.cpp
@@ -1,5 +1,6 @@
 #include <chrono>
 #include <future>
+#include <thread>
 
 #include <ros/connection.h>
 #include <ros/service_manager.h>
@@ -12,6 +13,10 @@ namespace foxglove_bridge {
 std::string retrieveServiceType(const std::string& serviceName, std::chrono::milliseconds timeout) {
   auto link = ros::ServiceManager::instance()->createServiceServerLink(serviceName, false, "*", "*",
                                                                        {{"probe", "1"}});
+
+  // Wait for connection to be ready
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
   if (!link) {
     throw std::runtime_error("Failed to create service link");
   } else if (!link->getConnection()) {
```

achim-k:
### Changelog
Fix node crashing in some cases when retrieving ROS1 service types

### Docs
None

### Description
In #306, it was reported that foxglove bridge crashed due to [this assertion](https://github.com/ros/ros_comm/blob/845f74602c7464e08ef5ac6fd9e26c97d0fe42c9/clients/roscpp/src/libros/connection.cpp#L274) when retrieving ROS1 service types. I believe that the reason for this assertion was that the header callback was set twice: Once in our code and once in the [ServiceServerLink](https://github.com/ros/ros_comm/blob/845f74602c7464e08ef5ac6fd9e26c97d0fe42c9/clients/roscpp/src/libros/service_server_link.cpp#L118). To avoid this, I have reworked the service retrieval to not create a `ServiceServerLink` instance but instead work with a raw `ros::Connection`. 

Fixes #306

### DIFF
--- a/ros1_foxglove_bridge/src/service_utils.cpp
+++ b/ros1_foxglove_bridge/src/service_utils.cpp
@@ -2,26 +2,46 @@
 #include <future>
 
 #include <ros/connection.h>
+#include <ros/connection_manager.h>
+#include <ros/poll_manager.h>
 #include <ros/service_manager.h>
-#include <ros/service_server_link.h>
+#include <ros/this_node.h>
+#include <ros/transport/transport_tcp.h>
 
 #include <foxglove_bridge/service_utils.hpp>
 
 namespace foxglove_bridge {
 
+/**
+ * Looks up the service server host & port and opens a TCP connection to it to retrieve the header
+ * which contains the service type.
+ *
+ * The implementation is similar to how ROS does it under the hood when creating a service server
+ * link:
+ * https://github.com/ros/ros_comm/blob/845f74602c7464e08ef5ac6fd9e26c97d0fe42c9/clients/roscpp/src/libros/service_manager.cpp#L246-L261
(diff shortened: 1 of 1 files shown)


achim-k:
@jurevreca12 confirmed that this patch fixes the issue, see https://github.com/foxglove/ros-foxglove-bridge/issues/306#issuecomment-2225421594

PR #327, opened 2024-11-19
Improve Error Reporting and Reduce Log Redundancy

rdumas01:
### Changelog
None

### Docs
None

### Description

The ROS 1 foxglove_bridge rejects parameter names with `.` and spams the terminal with redundant error messages.

While it is possible for ROS 1 parameters to have a `.` in their name, they are not "supposed" to (`Character [.] is not valid in Graph Resource Name`). For this reason, the foxglove_bridge will keep raising errors for parameter names with `.`

This PR addresses the error handling part of this issue: an unordered set tracks the encountered invalid parameters to throttle error log messages.

<table><tr><th>Before</th><th>After</th></tr><tr><td>

The bridge std_out gets periodically spammed with the same error messages until the invalid parameter is removed.
In the app, the only visible error message is too vague.

error_log_before

</td><td>

The bridge std_out only prints errors when invalid parameters are first encountered—one error message per invalid parameter.
In the app, the error message lists the invalid parameters.

error_log_after

</td></tr></table>

### DIFF
--- a/ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
+++ b/ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
@@ -4,7 +4,6 @@
 #include <mutex>
 #include <regex>
 #include <shared_mutex>
-#include <sstream>
 #include <string>
 #include <unordered_set>
 
@@ -49,6 +48,7 @@ constexpr uint32_t SUBSCRIPTION_QUEUE_LENGTH = 10;
 constexpr double MIN_UPDATE_PERIOD_MS = 100.0;
 constexpr uint32_t PUBLICATION_QUEUE_LENGTH = 10;
 constexpr int DEFAULT_SERVICE_TYPE_RETRIEVAL_TIMEOUT_MS = 250;
+constexpr int MAX_INVALID_PARAMS_TRACKED = 1000;
 
 using ConnectionHandle = websocketpp::connection_hdl;
 using TopicAndDatatype = std::pair<std::string, std::string>;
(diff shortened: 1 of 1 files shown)


ros1_foxglove_bridge/src/ros1_foxglove_bridge_nodelet.cpp
```diff
@@ -665,27 +667,49 @@ class FoxgloveBridge : public nodelet::Nodelet {
...
 
     if (!success) {
-      throw std::runtime_error("Failed to retrieve one or multiple parameters");
+      for (std::string& param : invalidParams) {
+        _invalidParams.insert(param);
+      }
+
+      if (!invalidParams.empty()) {
+        std::ostringstream errorMsg;
+        errorMsg << "Failed to retrieve the following parameters: ";
```
jtbandes, line 702: for something simple like this, I imagine you could use `+=` on a std::string rather than a stringstream, however this is fine too

jtbandes (approved):

jtbandes:
Looks good to me, might want to double-check with @achim-k if he has any feedback

defunctzombie:
If the issue is error logging spam - I'd propose using the throttled variants of these macros:

https://docs.ros.org/en/indigo/api/rosconsole/html/macros__generated_8h.html#a1c4cdfdfd4dcc53e6226f0533a0992b7

This avoids the needs to track the params. If you still want to track the params - maybe some upper bound on number of params tracked? I don't imagine too many problems with this vector growing but as written it is definitely a thing that can grow unbounded if you keep making invalid params.

achim-k (approved):

rdumas01:
> If the issue is error logging spam - I'd propose using the throttled variants of these macros

I think it's more a dedupe issue than a throttling one, the printing rate is already quite slow (basically it slowly floods the terminal with the same errors, and the more invalid parameters, the more error lines)

> maybe some upper bound on number of params tracked?

Yes, good point. I've never had projects with a lot of parameters, is 100 a realistic upper bound?

PR #347, opened 2025-04-28
Update default `asset_uri_allowlist` parameter to allow dashes

### LINKED ISSUE #346: Problem with allowed assets
First of all, I would like to congratulate you on a piece of code well done!

**Issue description:**
I've been playing around with Luxonis cameras and foxglove-bridge recently. I've noticed an unusual URDF error. After analyzing it, I've come to the source of the problem.

In the default launch, only paths containing only word characters, without dashes, are allowed. However, in the depthai_desctiption package, the model files contain underscores in the file name, which causes an error.

I suggest changing the default settings in the [launch file](https://github.com/foxglove/ros-foxglove-bridge/blob/main/ros2_foxglove_bridge/launch/foxglove_bridge_launch.xml#L18) from the current
```bash
  <arg name="asset_uri_allowlist"       default="['^package://(?:\\w+/)*\\w+\\.(?:dae|fbx|glb|gltf|jpeg|jpg|mtl|obj|png|stl|tif|tiff|urdf|webp|xacro)$']" />
```
to
```bash
  <arg name="asset_uri_allowlist"       default="'[^package://(?:[-\\w]+/)*[-\\w]+\\.(?:dae|fbx|glb|gltf|jpeg|jpg|mtl|obj|png|stl|tif|tiff|urdf|webp|xacro)$']" />
```
I'm not sure if this has any potential downsides, but it might save someone some debugging time in the future.

achim-k:
### Changelog
Update default `asset_uri_allowlist` parameter to allow dashes

### Docs
None

### Description
Updates the default `asset_uri_allowlist` parameter to support dashes in file / folder names. 

Fixes #346


achim-k:
Rolling CI jobs are failing due to undefined symbols in the fastcdr lib. 

```
/usr/bin/python3: symbol lookup error: /opt/ros/rolling/lib/libtest_msgs__rosidl_typesupport_fastrtps_c.so: undefined symbol: _ZN8eprosima7fastcdr3Cdr9serializeEPc
```

This should get fixed at some point when the `ros:rolling-ros-base` docker image gets updated.

See also https://github.com/eProsima/Fast-CDR/issues/266

jtbandes (approved):
Is the package name actually allowed to have dashes? I guess it's not harmful to allow it here, but not sure it's valid in ROS

Note to self: apparently `_` is included by default in `\w`

PR #353, opened 2025-05-16
Update asset_uri_allowlist regex to allow percent-encoding

tony2guo:
### Changelog
Update asset_uri_allowlist regex to allow percent-encoding

### Docs

None

### Description
This PR updates the default `asset_uri_allowlist` regex to properly support percent-encoded characters such as `%20` (whitespace) in file and folder names.

- Adds `%` to the regex to support whitespace and other percent-encoded characters in URIs.
- Ensures that dashes (`-`) are allowed in paths for ROS 1.
- Aligns ROS 1 behavior with the fix already applied in ROS 2 https://github.com/foxglove/ros-foxglove-bridge/pull/347.

Fixes https://github.com/foxglove/ros-foxglove-bridge/issues/346


achim-k (approved):
LGTM

Note that this change won't be released for ROS1 anymore: https://discourse.ros.org/t/preparing-for-final-noetic-sync-2025-05-29/43786

PR #355, opened 2025-06-10
Fix flaky publishing test

jtbandes:
### Changelog
None

### Docs

None

### Description

More follow-up from #354 

I believe the fact that a single FoxgloveBridge node is used for all tests was causing the PublisherTest to be flaky. I worked around the issue by using a unique topic name for each iteration of the test.

Some other tests seemed to be failing because the initial channels are sent before the test sets up its text message handler. I fixed that by moving the waitForChannel earlier.

Fixes FG-11910 / https://github.com/orgs/foxglove/discussions/1127 (again, hopefully)

### DIFF
--- a/ros2_foxglove_bridge/tests/smoke_test.cpp
+++ b/ros2_foxglove_bridge/tests/smoke_test.cpp
@@ -260,10 +260,13 @@ TEST(SmokeTest, testSubscriptionParallel) {
 
 TEST_P(PublisherTest, testPublishing) {
   const auto& [encoding, message] = GetParam();
+  // use a unique topic for each test to prevent tests from interfering with each other
+  static size_t testPublishingCount = 0;
+  ++testPublishingCount;
 
   foxglove::ClientAdvertisement advertisement;
   advertisement.channelId = 1;
-  advertisement.topic = "/foo";
+  advertisement.topic = "/testPublishing_" + std::to_string(testPublishingCount);
   advertisement.encoding = encoding;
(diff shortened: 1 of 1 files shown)


jtbandes:
@achim-k do you know why certain tests will sometimes take longer to run? For example, if you run `GTEST_REPEAT=10 ./build/foxglove_bridge/smoke_test --gtest_filter='*ExistingPublisherTestJSON/ExistingPublisherTest.testPublishingWithExistingPublisher*'` then the 1st iteration of the test sometimes has a delay (~0.5sec), but sometimes all the tests run without a visible delay. i.e. the output of `time smoke_test ...` for this specific test is very bimodal. Is this delay just coming from somewhere in ROS?

defunctzombie (approved):

achim-k (approved):

achim-k:
> @achim-k do you know why certain tests will sometimes take longer to run? For example, if you run `GTEST_REPEAT=10 ./build/foxglove_bridge/smoke_test --gtest_filter='*ExistingPublisherTestJSON/ExistingPublisherTest.testPublishingWithExistingPublisher*'` then the 1st iteration of the test sometimes has a delay (~0.5sec), but sometimes all the tests run without a visible delay. i.e. the output of `time smoke_test ...` for this specific test is very bimodal. Is this delay just coming from somewhere in ROS?

No idea. What you describe reminded me a little of the ROS2 daemon, which gets started the first time you use a ros2 cli command (e.g. ros2 topic list), but I believe that that daemon is only used for that.

PR #360, opened 2025-06-26
Run ros2 executable without component manager

achim-k:
### Changelog
Fix warning when running ROS2 executable with custom node name

### Docs
None

### Description
Changes the ROS2 executable to use the standalone foxglove_bridge node rather than using a component manager and loading the foxglove_bridge component in it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -186,6 +186,7 @@ elseif("$ENV{ROS_VERSION}" STREQUAL "2")
     )
     target_include_directories(foxglove_bridge SYSTEM PRIVATE ${rclcpp_INCLUDE_DIRS})
     target_link_libraries(foxglove_bridge
+      foxglove_bridge_component
       rclcpp::rclcpp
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -186,6 +186,7 @@ elseif("$ENV{ROS_VERSION}" STREQUAL "2")
     )
     target_include_directories(foxglove_bridge SYSTEM PRIVATE ${rclcpp_INCLUDE_DIRS})
     target_link_libraries(foxglove_bridge
+      foxglove_bridge_component
```
jtbandes, line 189: what's this for?

achim-k:
> Was there any particular reason we used component manager in the first place?

I don't remember. Maybe for testing if component loading works correctly? :man_shrugging:

CMakeLists.txt
achim-k: Required to set the correct include path for the added header include

PR #362, opened 2025-06-27
Set up build files and working copy of ros2_foxglove_bridge for SDK

ericmlujan:
### Changelog
None

### Docs
None

### Description
Make a copy of `ros2_foxglove_bridge/` in the `ros2_foxglove_bridge_sdk/` directory, and introduce a build switch (`USE_FOXGLOVE_SDK`) to determine which will be used as the source directory for building the `foxglove_bridge_component` target. There are also changes to the CMakeLists.txt file to enable fetching the C++ SDK and configuring it as a link dependency if `USE_FOXGLOVE_SDK` is enabled.

This is intended to create a workspace for developing the SDK version of the bridge in `main` with incremental and easily-reviewable PRs and without needing to maintain a long-lived feature integration branch. Isolating the source files and build flags should allow for this without affecting production builds.

Tested that the bridge builds as normal if `USE_FOXGLOVE_SDK` is set to false, and was also able to successfully build in Docker with the SDK enabled (i.e. `make USE_FOXGLOVE_SDK=ON jazzy`).

Fixes: FG-12086

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
       matrix:
         ros_distribution: [jazzy, kilted, rolling]
 
-    name: Test (ROS ${{ matrix.ros_distribution }}, (SDK Server)
+    name: Test (ROS ${{ matrix.ros_distribution }}, SDK Server)
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4


ericmlujan:
Note to reviewers that fe2c39f is a mechanical folder copy. The only substantive changes in this PR are in build and Docker files

CMakeLists.txt
```diff
@@ -15,12 +15,15 @@ set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 macro(enable_strict_compiler_warnings target)
-  if (MSVC)
-    target_compile_options(${target} PRIVATE /WX /W4)
-  elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-    target_compile_options(${target} PRIVATE -Wall -Wextra -Wpedantic -Werror -Wold-style-cast -Wfloat-equal -Wmost -Wunused-exception-parameter)
-  else()
-    target_compile_options(${target} PRIVATE -Wall -Wextra -Wpedantic -Werror -Wold-style-cast -Wfloat-equal)
+  # Disable warnings if USE_FOXGLOVE_SDK is enabled until SDK with fixed warnings is released
```
jtbandes, line 18: 🤔   not sure what the warning issues you ran into were, but if the SDK is added to include directories with the SYSTEM option, that may silence warnings from the SDK's own headers (although we should still try and fix those!)

CMakeLists.txt
```diff
@@ -51,6 +56,53 @@ if (ENDIAN)
...
+  file(GLOB_RECURSE _foxglove_sdk_sources ${foxglove_sdk_SOURCE_DIR}/src/*.cpp)
+  add_library(foxglove_sdk SHARED
+    ${_foxglove_sdk_sources}
+  )
+  file(GLOB_RECURSE _foxglove_sdk_headers ${foxglove_sdk_SOURCE_DIR}/include/*.hpp)
+  target_include_directories(foxglove_sdk
+    PUBLIC
+      $<BUILD_INTERFACE:${foxglove_sdk_SOURCE_DIR}/include>
+      $<INSTALL_INTERFACE:include>
+  )
```
jtbandes, line 90: wonder if we should try to make this match closer with https://docs.foxglove.dev/docs/sdk -- or if there were reasons that didn't work?  (although I like your GLOB_RECURSE better than the GLOB in docs...maybe we should update the docs?)

CMakeLists.txt
```diff
@@ -51,6 +56,53 @@ if (ENDIAN)
...
+  install(FILES ${_foxglove_sdk_headers}
+    DESTINATION include/${PROJECT_NAME}
+  )
+  install(TARGETS foxglove_sdk
+    EXPORT foxglove_sdk
+    ARCHIVE DESTINATION lib
+    LIBRARY DESTINATION lib
+    RUNTIME DESTINATION bin
+  )
```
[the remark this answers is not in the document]
ericmlujan: This is a leftover from when I was using a dynamic library version and not static linkage. Removed in 9a92585.

jtbandes (approved):
Seems good to me, as long as @achim-k is aligned on the approach of having side-by-side implementations until the SDK one is ready

ericmlujan:
@jtbandes thanks again for the review, agree that @achim-k should have the last word here :)

jtbandes:
ooh, can we also add this to CI so we can get the tests up and running?  (not sure we need to duplicate the entire build matrix for all ros distros, but maybe start with one or two at least?)

ericmlujan:
@jtbandes 👍 Addressed CI setup request in f8b5fc7

achim-k (approved):
> Seems good to me, as long as @achim-k is aligned on the approach of having side-by-side implementations until the SDK one is ready

No objections from my side

ericmlujan:
An FYI to reviewers that the intermittently failing unit test is also present on `main` and tracked in FG-12227

PR #364, opened 2025-07-08
Implement server and client message publishing using the SDK server implementation

ericmlujan:
### Changelog
None

### Docs
None

### Description
Use the Foxglove SDK to implement ROS message passing from the server to connected WebSocket clients. 

Implement capability to listen to advertisements of channels from connected clients and pass them through to ROS.

This branch currently depends on changes in foxglove/foxglove-sdk that are not yet part of a released artifact. Currently have built artifacts temporarily checked in as LFS objects for development, but that will be reversed once a release is made.

Note that some of the existing ws-protocol code still exists in commented-out form for my reference as I develop. I plan on cleaning this up incrementally as the majority of the code in `ros_foxglove_bridge.cpp` becomes SDK-ified.

Existing suite of server publication and client advertising unit tests pass, including cases that test latched topic functionality. Verified that an existing mcap file from nuScenes can be played and viewed in the Foxglove webapp using `ros2 bag play`. 

https://github.com/user-attachments/assets/75348110-439b-4295-a6d2-cb0056af2124

Fixes: FG-12233
Fixes: FG-12282

### DIFF
--- a/ros2_foxglove_bridge_sdk/include/foxglove_bridge/ros2_foxglove_bridge.hpp
+++ b/ros2_foxglove_bridge_sdk/include/foxglove_bridge/ros2_foxglove_bridge.hpp
@@ -75,8 +75,9 @@ private:
 
   // BEGIN New SDK Components
   std::unique_ptr<foxglove::WebSocketServer> _sdkServer;
-  std::unordered_map<uint64_t, foxglove::RawChannel> _sdkChannels;
+  std::unordered_map<ChannelId, foxglove::RawChannel> _sdkChannels;
   std::unordered_map<ChannelAndClientId, Subscription, PairHash> _sdkSubscriptions;
+  std::unordered_map<ChannelAndClientId, ClientAdvertisement, PairHash> _clientAdvertisedTopics;
   // END New SDK Components
 
   std::unique_ptr<foxglove_ws::ServerInterface<ConnectionHandle>> _server;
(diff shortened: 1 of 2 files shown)


ros2_foxglove_bridge_sdk/include/foxglove_bridge/ros2_foxglove_bridge.hpp
```diff
@@ -67,7 +89,8 @@ class FoxgloveBridge : public rclcpp::Node {
   std::unordered_map<foxglove_ws::ChannelId, foxglove_ws::ChannelWithoutId> _advertisedTopics;
   std::unordered_map<foxglove_ws::ServiceId, foxglove_ws::ServiceWithoutId> _advertisedServices;
   std::unordered_map<foxglove_ws::ChannelId, SubscriptionsByClient> _subscriptions;
-  PublicationsByClient _clientAdvertisedTopics;
+  std::unordered_map<std::pair<uint32_t, uint32_t>, ClientAdvertisement, PairHash>
```
achim-k, line 92: what is stored inside the pair? It would be more readable if you would use the type aliases defined above

ros2_foxglove_bridge_sdk/include/foxglove_bridge/ros2_foxglove_bridge.hpp
```diff
@@ -90,15 +113,16 @@ class FoxgloveBridge : public rclcpp::Node {
 
   void subscribeConnectionGraph(bool subscribe);
 
-  void subscribe(foxglove_ws::ChannelId channelId, ConnectionHandle clientHandle);
+  void subscribe(uint64_t channelId, const foxglove::ClientMetadata& client);
```
achim-k, line 116:
```suggestion
  void subscribe(ChannelId channelId, const foxglove::ClientMetadata& client);
```

also below

ros2_foxglove_bridge_sdk/include/foxglove_bridge/ros2_foxglove_bridge.hpp
achim-k: ah, I guess this should be 
```suggestion
  std::unordered_map<ChannelAndClientId, ClientAdvertisement, PairHash>
```
?

PR #373, opened 2025-07-24
Decouple SDK implementation from foxglove_bridge_base

ericmlujan:
### Changelog
None

### Docs
None

### Description
Remove dependencies between `ros2_foxglove_bridge_sdk` and the `foxglove_bridge_base` library.

A couple required files (MessageDefinitionCache and the common.hpp header) were copied over into the `ros2_foxglove_bridge_sdk` tree. I also revisited the version header generation code.

The most involved change in this PR was distilling out the minimum amount of code from the old websocket_client code to support running the tests in smoke_test. Functions were largely copied verbatim, but I did away with the ClientInterface base class (since we'll only be using the client for testing going forward), and I implemented my own stub structs and utilities for wrapping, serializing, and deserializing the WebSocket messages that are used in testing (see protocol_types.hpp). The exception were Parameters, where I'm using the Foxglove SDK types directly, as re-implementing them fully felt unnecessary; I only wrote functions for serializing/deserializing them to/from JSON.

I also moved bridge code that previously was in the `foxglove` namespace (which is also used by the SDK) to the `foxglove_bridge` namespace, to make it clearer what is bridge-specific code.

<table><tr><th>Before</th><th>After</th></tr><tr><td>

`ros2_foxglove_bridge_sdk` targets depend on `foxglove_bridge_base`

</td><td>

`ros2_foxglove_bridge_sdk` and tests depend only on external libraries like `foxglove_sdk`, OpenSSL, etc.

</td></tr></table>

Fixes: FG-12451


ros2_foxglove_bridge_sdk/include/foxglove_bridge/test/protocol_types.hpp
```diff
@@ -0,0 +1,89 @@
...
+  CallId callId;
+  std::string encoding;
+  std::vector<std::byte> data;
+};
+
+struct ServiceResponse {
+  ServiceId serviceId;
+  CallId callId;
+  std::string encoding;
+  std::vector<uint8_t> data;
```
bryfox, line 71: Just curious, why not `std::byte` here too?

bryfox (approved):

ros2_foxglove_bridge_sdk/include/foxglove_bridge/test/protocol_types.hpp
ericmlujan: I was avoiding a cast in an earlier implementation of the logic; I changed my approach and made it compatible with std::byte and forgot to update this 🙃

PR #375, opened 2025-07-25
Remove Iron from CI

jtbandes:
### Changelog
Removed Iron builds from CI since Iron reached EOL in Dec 2024.

### Docs

None

### Description

There was an intermittent test failure (segfault) that is only reproducible in Iron. Since Iron is already EOL, there's no need to bother trying to fix the test.

- Remove from CI build matrix (also removed the required check in repo settings)
- Remove from Makefile
- Remove badge from README (or we could leave it there? 🤷)

Resolves FG-12227

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


ericmlujan (approved):
🔪

PR #379, opened 2025-10-01
Update the README to only talk about ROS 1.

clalancette:
### Changelog

All ROS 2 development has moved to
https://github.com/foxglove/foxglove-sdk, so make that clear in the README.

### Docs

This is the update to the docs :).

### Description

Update the docs to only mention ROS 1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ Parameters are provided to configure the behavior of the bridge. These parameter
 ### Fetch source, install dependencies, and build for ROS 1
 
 ```bash
+source /opt/ros/noetic/setup.bash
 mkdir -p foxglove_bridge_ws/src
 cd foxglove_bridge_ws
 git clone https://github.com/foxglove/ros-foxglove-bridge.git src/ros-foxglove-bridge
@@ -66,7 +67,6 @@ rosdep update
 rosdep install --ignore-src --default-yes --from-path src
 catkin_make install
 source install/setup.bash
-roslaunch --screen foxglove_bridge foxglove_bridge.launch
 ```
 
 ## Clients
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ The `foxglove_bridge` uses the [Foxglove WebSocket protocol](https://github.com/
 
 ## Installation
 
-**Note**: While binary packages are available for ROS 1, all ROS 1 releases are End-of-Life. That means that new binary packages of `foxglove_bridge` cannot be released into those releases, and the versions are quite outdated. It is highly recommended to [build foxglove_bridge from source](#building-from-source) for the latest bug fixes. For similar reasons, `foxglove_bridge` is in maintenance mode and only bug fixes will be applied.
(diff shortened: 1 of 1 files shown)


achim-k (approved):

README.md
```diff
@@ -1,30 +1,18 @@
...
 ## Installation
 
-**Note**: This project is under active development and binary releases of `foxglove_bridge` might be quite outdated. For the latest features and bug fixes, consider [building foxglove_bridge from source](#building-from-source).
-
-The `foxglove_bridge` package is available for ROS 1 Melodic and Noetic, and ROS 2 Humble and Rolling. Earlier releases of ROS will not be supported due to API design and/or performance limitations. The package can be installed with the following command:
-
-```bash
-sudo apt install ros-$ROS_DISTRO-foxglove-bridge
-```
+**Note**: While binary packages are available for ROS 1, all ROS 1 releases are End-of-Life. That means that new binary packages of `foxglove_bridge` cannot be released into those releases, and the versions are quite outdated. It is highly recommended to [build foxglove_bridge from source](#building-from-source) for the latest bug fixes. For similar reasons, `foxglove_bridge` is in maintenance mode and only bug fixes will be applied.
```
amacneil:
```suggestion
**Note**: While binary packages are available for ROS 1, all ROS 1 releases are End-of-Life. That means that new binary packages of `foxglove_bridge` cannot be released into those releases, and the versions are quite outdated. It is highly recommended to [build foxglove_bridge from source](#building-from-source) for the latest bug fixes. For similar reasons, this `foxglove_bridge` package for ROS 1 is in maintenance mode and only bug fixes will be applied.
```

README.md
```diff
@@ -102,45 +67,14 @@ rosdep install --ignore-src --default-yes --from-path src
 
 ### ROS 1
 ```
-catkin_make
-source install/local_setup.bash
+catkin_make install
```
amacneil, line 68: I think it would be worth adding git checkout instructions here.

like if I had an existing ros 1 workspace, is there a recommended place for where to check out this repo to? would I still need to `source install/setup.bash` if I have an existing ros workspace? how would I add this to an existing launch file?

some of this stuff might be obvious to ros veterans but its helpful to make the install steps self contained (I feel the same way about our ros 2 source install docs - would be nice to update those in the sdk repo too)

README.md
clalancette: Updated in 5f7d092

README.md
clalancette: The instructions were there, but were in the above block.  Given that this is only for ROS 1 now, I actually combined the two blocks and made it a more end-to-end example of creating a workspace, checking out this repository, building with catkin_make, and sourcing the install space.  Let me know what you think.

ericmlujan (approved):

clalancette:
I'm going to go ahead and merge this one.  If there are additional instructions we want to add/clarify, we can always do a follow-up PR.

PR #380, opened 2025-10-02
Remove all references to ROS 2.

clalancette:
### Changelog

All ROS 2 development has moved to https://github.com/foxglove/foxglove-sdk.  This repository now only holds the legacy ROS 1 foxglove_bridge.

### Docs

The associated changes to the README are happening over in #379 

### Description

All ROS 2 development has now moved to
https://github.com/foxglove/foxglove-sdk, so remove all of the ROS 2 code from here.

I did basic smoke testing with ROS 1 before and after this change, and ROS 1 continues to work from this repository.


.github/workflows/ci.yml
```diff
@@ -6,22 +6,11 @@ on:
...
-    strategy:
-      fail-fast: false
-
-    name: Lint
-    runs-on: ubuntu-latest
-
-    steps:
-      - uses: actions/checkout@v4
-      - run: make lint
-
```
clalancette, line 19: Note to reviewers: I intentionally removed the linting workflow here, since it was depending on the ROS 2 devcontainer to work.  While not having linting on this repository is unfortunate, given that this is in maintenance mode I don't think we'll have a huge amount of change, and thus I think it is acceptable to go without linting here.

ericmlujan: Sounds reasonable. I filed [FLE-94](https://linear.app/foxglove/issue/FLE-94/re-introduce-lint-to-ros1-foxglove-bridge) so we have the idea of re-adding lint tracked somewhere, but it's not a high priority at all.

ericmlujan (approved):
LGTM, let's ship it 🔪🫡

